PR #3, opened 2012-08-23
Added empty directories and support for Maven users with non-standard local repo location

eddgrant:
Hi all,

These commits add some empty directories (which were causing the Maven build to fail) which I think were deleted (Git doesn't support empty directories) when the apache svn repo was moved in to GitHub. I have re-created them each containing a single, empty .gitignore file which allows Git to persist them.

Secondly I have changed the way the local Maven repo is looked up, it was previously hard coded to look in the default location, it now takes the value from the user's settings.xml so it can pick up an overridden value if there is one, else will still use the default if not set in settings.xml

Hope this is helpful - a 'mvn clean install' now completes successfully from a fresh GitHub fork without it being necessary to make any modifications which should be encouraging to new roller users.

Cheers,

Edd


lbtc-xxx:
Hi Edd, Thanks for the PR and sorry to have kept you waiting for long time.

It looks like Apache Roller team seldom check GitHub. I've filed the one causing the Maven build to fail https://issues.apache.org/jira/browse/ROL-2086 , And I would start working to fix it.

As for making use `${settings.localRepository}`, The relevant part has gone in the time of merging Maven projects so it's unnecessary now.

eddgrant:
Hey,

No worries, thanks for getting back to me and for raising ROL-2086.

Cheers.

Edd

lbtc-xxx:
Hi Edd,

I've just committed the fix of `ROL-2086` at commit 38e9c3f029b4d053dc7e852c47f4d6eb081166d5 . Now you can build Roller using distribution that has been obtained from GitHub.

Thank you again for the contribution,

Kohei

PR #11, opened 2018-04-14
Upgrade Xinha to the latest

nuzayats:
This upgrades the Xinha HTML editor to the latest (1.5.1). http://trac.xinha.org/wiki/ReleaseNotes#Xinha1.5.1

The directory where there are the Xinha files are replaced by the latest Xinha distribution (other than the contrib directory, which contains some unnecessary files like something.php). The path in EntryEditor.jsp is updated to use the new version.

I've quickly checked on some Roller deployments and it seems to be working fine.


lbtc-xxx:
A good thing of the new version is that now it works on iOS Safari.

snoopdave:
Looks good to me. Kind of amazing how many files Xinha has!

PR #28, opened 2019-06-03
Improved: Editor should confirm user before leaving the page (ROL-2136)

adityasharma7:
On the pages where editors are used, if the user has done some unsaved changes. Before closing window/tab it should confirm the user.
Done the following changes:
1. Included UnsavedChanges plugin for Xinha editor that implicitly handles it
2. Added custom code to add confirmation box for text area which adds an event handler to beforeunload event if the text area is changed.

### DIFF
--- a/app/src/main/webapp/WEB-INF/jsps/editor/EntryEditor.jsp
+++ b/app/src/main/webapp/WEB-INF/jsps/editor/EntryEditor.jsp
@@ -127,7 +127,6 @@
         }
         // Added event listener to confirm once the editor content is changed
         jQuery("#edit_content").one("change", function() {
-            console.log("this");
             jQuery(window).on("beforeunload", function(event) {
                 // Chrome requires returnValue to be set and original event is found as originalEvent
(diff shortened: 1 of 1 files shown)


snoopdave:
Looks good to me.

adityasharma7:
Thanks @snoopdave for the review. 
I was just wondering we are already working on bootstrap-ui where Xinha editor will be replacing summernote. We should merge the PR only if we are planning for a release before bootstrap changes gets merged. 
In case we are planning next release to be the one with bootstrap-ui only I will create a new PR for the bootstap-ui branch having these changes specific to summernote and we will close this one.  WDYT?

snoopdave:
@adityasharma7 good point. I think this change should be merged to the roller-5.2.x branch so that it is included in any figure 5.2.x releases.

You could also merge it to master, but I hope to merge bootstrap-ui soon, and you're right, Summernote will need a different fix.

adityasharma7:
I have already created a separate PR for 5.2.x https://github.com/apache/roller/pull/30 and will create a new one for bootstrap-ui branch hence closing it.
Thanks @snoopdave

PR #29, opened 2019-06-12
Fixed: NOTICE file does not have standard content (ROL-2138)

adityasharma7:
1. Rearranged the content as per the standard for Notice and License files
2. Removed credits from Notice file


snoopdave:
Looks good to me. Please merge when ready.

PR #34, opened 2019-07-17
Apache Lucene Renovation

mbien:
Hello!

updated apache lucene from 4.x to 8.1.x which required some code changes. Since lucene can only read indexes which are not older than 1-2 major versions (according to the exception i got), I added a simple check which causes a rebuild task if the index is unreadable. (first commit)

The second commit removes usages of the deprecated RAMDirectory. I decided to not migrate it to MMapDirectory because the code path isn't used anywhere (or tested). Furthermore: MMapDirectory extends FSDirectory (which is currently used), so it wouldn't use the old code in any case.

tests are green and i also deployed a build to my blog and tested the index rebuilding once.

- - - - 
the only change i am not sure about is: https://github.com/mbien/roller/commit/10370c129dd5d5cb906095fc23be41bb69014179#diff-afa10aab8f67b8be187418bf4264950dR140

Since the search results are sorted by pubdate, lucene 8 wants this field to be a DocValue of type SORTED instead of NONE and throws a IllegalStateException. So i changed it to SortedDocValuesField which made lucene happy. This field type does not have the stored yes/no property which the StringField had. Javadoc recommends to add another StoredField if storage is required - I just don't know if its needed for roller so I left it out.

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -49,7 +49,7 @@ limitations under the License.
         <guice.version>4.2.2</guice.version>
         <log4j.version>1.2.17</log4j.version>
         <log4j2.version>2.10.0</log4j2.version>
-        <lucene.version>4.10.4</lucene.version>
+        <lucene.version>8.1.1</lucene.version>
         <oauth-core.version>20100527</oauth-core.version>
         <maven-war.version>3.1.0</maven-war.version>
         <maven-surefire.version>2.17</maven-surefire.version>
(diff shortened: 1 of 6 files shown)


snoopdave:
@mbien Wow this looks great. I will review and hope to merge this week.

mbien:
let me know if you need any changes or a rebase etc

snoopdave:
Thanks @mbien!

PR #35, opened 2019-08-12
fix global config jsp wrong tag

zorrofox:
In the [WEB-INF/jsps/admin/GlobalConfig.jsp](https://github.com/apache/roller/blob/83d8f8bb4ca12e1a1000fcd8e3422de3f046fcb3/app/src/main/webapp/WEB-INF/jsps/admin/GlobalConfig.jsp#L43) file line 43:
`                          list="weblogs" listKey="name" listValueKey="handle"/>` is not correct, should be miss coding.


zorrofox:
I found more bugs in this page, and will raise a new PR. Thanks!

PR #37, opened 2019-09-01
Upgrade jQuery to 3.4.1  ROL-2150

adityasharma7:


snoopdave:
This a a good idea to update jQuery and the right way to do it is by changing the Webjars dependency the Roller app's pom.xml file. This is what is in there now:

`        <dependency>
            <groupId>org.webjars</groupId>
            <artifactId>jquery</artifactId>
            <version>3.3.1</version>
        </dependency>

        <dependency>
            <groupId>org.webjars</groupId>
            <artifactId>jquery-ui</artifactId>
            <version>1.12.1</version>
        </dependency>`

The webjars dependecies are imported into the web pages via head.jsp.

We should remove the jQuery files from Roller's repo by removing these direrectories
`app/src/main/webapp/roller-ui/js/jquery*` and we should remove the references to that directory in all of the JSP files that you changed.  All of those files should instead get the jQuery dependency from head.jsp which is included in all JSP files (via Struts / Tiles).

This will make it easier to upgrade jQuery in the future because we will only need to change `pom.xml` and `head.jsp`.

adityasharma7:
Thanks @snoopdave for the heads up :)
 
I traced back, Js libraries import using webjars dependencies is available in master[1] & roller 6[2] but not in 5.2.x. [3] [4]

Actually I thought to make the change as per the current code of the roller-5.2.x branch.
Though it will be a good idea to get those changes to 5.2.x.
Wdyt?


1. https://github.com/apache/roller/blob/master/app/pom.xml
2. https://github.com/apache/roller/blob/roller-6.0.x/app/pom.xml
3. https://markmail.org/message/zslwmmy5is23psc4
4. https://github.com/apache/roller/blob/roller-5.2.x/app/pom.xml

snoopdave:
Yes, I think it would be good to get those changes into 5.2.x as well and I'd be willing to help out with another 5.2.x release when we're ready. It's up to you if you want to add webjars to 5.2.x or do it the old way.

adityasharma7:
Sure Dave :)

adityasharma7:
Apologies for the delay. I got engaged with some unavoidable priorities.
Following MR contains the following changes:
1. Used web jars for including jQuery and jQuery UI
2. jQuery updated to latest 3.4.1
Thanks @SanchitAgnihotri for your efforts!
3. jQuery UI updated to latest 1.12.1
Thanks @anushigupta for your efforts!
4. Xinha editor updated to the latest 1.5.3

Additionally, FixedTypeError on Media File Add page

TypeError: fileControls.size is not a functionmediaFileAdd.rol:283:42    <anonymous> http://localhost:8080/roller/roller-ui/authoring/mediaFileAdd.rol?weblog=test&directoryName=default:283

I will wait for a day before I merge.

PR #42, opened 2019-10-13
Improved: Editor should confirm user before leaving the page (ROL-2136)

adityasharma7:


adityasharma7:
Need to implement a different solution for the master and 6.0 as it is having Summernote editor

PR #46, opened 2019-11-03
Fixed: Variables should be declared explicitly in Custom JS code (ROL-2157)

adityasharma7:
Pattern is identified and reported at  sonacloud.io as Blocker

JavaScript variable scope can be particularly difficult to understand and get right. The situation gets even worse when you consider the accidental creation of global variables, which is what happens when you declare a variable inside a function or the for clause of a for-loop without using the let, const or var keywords.


snoopdave:
Looks good to me.

adityasharma7:
Thanks @snoopdave for the review

PR #48, opened 2019-11-07
Improved: Replaced pom.basedir deprecated property with project.basedir. (ROL-2158)

isdixit:
NOTE:
Deprecated features are those that have been retained temporarily for backward compatibility, but which will eventually be removed. In effect, deprecation announces a grace period to allow the smooth transition from the old features to the new ones. In that period, no use of the deprecated features should be added, and all existing uses should be gradually removed.


adityasharma7:
Thanks @isdixit!

PR #51, opened 2019-11-23
made it easier to run roller on latest JVMs (tested with JDK 13).

mbien:
- use "no_aop" version of guice which doesn't bundle its own ASM version
- bumped ASM dependency to 7.2 for Java 13 compatibility

fixed some jetty warnings by not adding the javax.activation api classes twice.
minor maven dependency updates.
added version checker maven plugin, can be run with: mvn versions:display-dependency-updates.

### DIFF
--- a/version-rules.xml
+++ b/version-rules.xml
@@ -1,0 +1,16 @@
+<ruleset comparisonMethod="maven"
+         xmlns="http://mojo.codehaus.org/versions-maven-plugin/rule/2.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://mojo.codehaus.org/versions-maven-plugin/rule/2.0.0 http://mojo.codehaus.org/versions-maven-plugin/xsd/rule-2.0.0.xsd">
+    <ignoreVersions>
+        <!--only release builds-->
+        <ignoreVersion type="regex">.*(alpha|ALPHA|beta|BETA|b|rc|RC|M).*</ignoreVersion>
+        
+        <!--ignore some specific version strings which don't follow the common version format-->
+        <ignoreVersion type="exact">20020829</ignoreVersion>
+        <ignoreVersion type="exact">20040616</ignoreVersion>
+        <ignoreVersion type="exact">20040102.233541</ignoreVersion>
+        <ignoreVersion type="exact">20031027.000000</ignoreVersion>
+        <ignoreVersion type="exact">20030418.083655</ignoreVersion>
+    </ignoreVersions>
+
+</ruleset>
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -40,21 +40,22 @@ limitations under the License.
         <jstl.version>1.2</jstl.version>
 
         <angular.version>1.7.8</angular.version>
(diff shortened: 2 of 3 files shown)


snoopdave:
Overall this looks good and I'm always happy to see dependencies upddated. 

Happy to see that this allows Roller to run on JDK 13. Does it also work with JDK11?

version-rules.xml
```diff
@@ -0,0 +1,16 @@
+<ruleset comparisonMethod="maven"
+         xmlns="http://mojo.codehaus.org/versions-maven-plugin/rule/2.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://mojo.codehaus.org/versions-maven-plugin/rule/2.0.0 http://mojo.codehaus.org/versions-maven-plugin/xsd/rule-2.0.0.xsd">
+    <ignoreVersions>
+        <!--only release builds-->
+        <ignoreVersion type="regex">.*(alpha|ALPHA|beta|BETA|b|rc|RC|M).*</ignoreVersion>
+        
+        <!--ignore some specific version strings which don't follow the common version format-->
+        <ignoreVersion type="exact">20020829</ignoreVersion>
```
snoopdave: What is the purpose of these ignoreVersions?

mbien: version-rules.xml is for the version-maven-plugin* i added to the main pom.xml.

the plugin can be run manually with "mvn versions:display-dependency-updates" and will list all dependencies in this project which could be updated (just for convenience, the command itself does not update anything - but it could be used for that too).

The ignoreVersion regex tries to filter out any recommendations which aren't release versions. The remaining rules in that file blacklist some old artifact snapshots.

the maven-antrun-plugin for example has an old artifact with the version 20020829, which would be always recommended by the plugin if not blacklisted since it would read it as very high version number. There might be a better way of doing this but luckily it needed only a handful of entries on the blacklist.

*https://www.mojohaus.org/versions-maven-plugin/index.html

mbien:
> Happy to see that this allows Roller to run on JDK 13. Does it also work with JDK11?
tests are green on AdoptOpenJDK jdk-11.0.5+10 if I ignore SubscriptonFunctionalTest:testSubscriptionLookups which never passed for me :)

mbien:
My blog is getting only a few remaining ASM related exceptions from struts2 classloader on startup on JDK 13 + latest jetty. Didn't manage to get rid of them yet but they don't seem to influence anything. 

Its surprising to me to see so many libraries which try to bundle their own version of ASM.

snoopdave:
I've tested Roller master with the new version of Struts (2.5.22), which is coming out soon, I no longer see the ASM errors on JDK 11. I have not tried 13 yet.

mbien:
> 
> 
> I've tested Roller master with the new version of Struts (2.5.22), which is coming out soon, I no longer see the ASM errors on JDK 11. I have not tried 13 yet.

awesome! I am going to try it out later and report back if it solved the issue with 13 too.

mbien:
deployed a build with struts 2.5.22 from apache staging on JDK 13 with language level 13 + latest jetty and got no exceptions. Going to keep this build running.

PR #69, opened 2020-10-19
Fix for dusty integration tests

mbien:
integration tests should be working again.

commit msg:
 - the pom was broken possibly from a merge conflict from some time ago
 - fixed some paths and dependencies
 - adapted for firefox driver api changes
 - some tests needed minor updates due to the new bootstrap UI
 - some tests threw StaleElementReference exceptions when they tried to access elements after
   clicking a hyperlink which reloaded the page; fixed that too

tested with:
 cd it-selenium
 mvn clean integration-test -Dwebdriver.gecko.driver=/path/to/geckodriver

used:
 geckodriver v0.26.0 on 64bit linux, latest firefox, jdk 14 and everything worked fine.

note:
 do a mvn install of the roller root project first. Doing this for ./app is not sufficient
 since the tests look for the project pom/war in local repo, not roller.war.
note2:
 the test will display a few meters of jetty "class scanned from multiple locations" warnings
 since the classpath is apperently set to target/ and the local copy in the maven repo.
 Isn't convention over configuration great? I did not further attempt to fix this after trying
 out some different dependency scopes without any effect.

### DIFF
--- a/it-selenium/pom.xml
+++ b/it-selenium/pom.xml
@@ -48,7 +48,7 @@
         <dependency>
             <groupId>org.seleniumhq.selenium</groupId>
             <artifactId>selenium-java</artifactId>
-            <version>3.4.0</version>
+            <version>3.141.59</version>
             <exclusions>
                 <exclusion>
                     <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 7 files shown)


mbien:
@snoopdave forgot to ping :-)

PR #72, opened 2020-12-30
Upgrade to Jetty 11 maven plugin

snoopdave:
Also some attempts to reduce number of warnings about classes loaded from multiple locations, and some import cleanups. This change also makes it possible to run `mvn jetty:run` from the top-level directory and not only in the `app` module.

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -55,7 +55,6 @@ limitations under the License.
         <maven-antrun.version>1.0b3</maven-antrun.version>
         <rome.version>1.13.1</rome.version>
         <slf4j.version>1.7.30</slf4j.version>
-        <spring.version>5.2.7.RELEASE</spring.version>
         <spring.security.version>5.3.3.RELEASE</spring.security.version>
         <struts.version>2.5.22</struts.version>
         <velocity.version>2.2</velocity.version>
@@ -87,6 +86,7 @@ limitations under the License.
             <groupId>javax.servlet</groupId>
             <artifactId>jstl</artifactId>
             <version>${jstl.version}</version>
+            <scope>provided</scope>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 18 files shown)


pom.xml
```diff
@@ -25,10 +25,6 @@ limitations under the License.
     <version>6.1.0-SNAPSHOT</version>
     <packaging>pom</packaging>
 
-    <prerequisites>
```
snoopdave, line 28: This prerequisites tag is only to be used with Maven Plugin project POMs, not plain old modules like Roller's.

mbien:
@snoopdave did you mean Jetty 10? Jetty 11 runs Jakarta EE 9 and most dependencies are not ready for that yet.

app/pom.xml
```diff
@@ -55,7 +55,6 @@ limitations under the License.
         <maven-antrun.version>1.0b3</maven-antrun.version>
         <rome.version>1.13.1</rome.version>
         <slf4j.version>1.7.30</slf4j.version>
-        <spring.version>5.2.7.RELEASE</spring.version>
```
mbien, line 58: was this intended? spring and spring sec are now defined in two different poms. (asking before i rebase)

does this help to reduce selenium warnings or has it other reasons?

snoopdave:
> @snoopdave did you mean Jetty 10? Jetty 11 runs Jakarta EE 9 and most dependencies are not ready for that yet.

Yes. Thanks! I edited the title to fix that.

app/pom.xml
snoopdave: I was trying to use dependencyManagement in the parent POM to force all Spring deps to the same version, but I don't think it really helped with all those warnings you see when Jetty starts in app or it-selenium. 

I'm going to back this change out.

PR #77, opened 2021-02-24
Fixed: type specification when using generics

ymaheshwari1:
Java 7 introduced the diamond operator (<>) to reduce the verbosity of generics code. For instance, instead of having to declare a List's type in both its declaration and its constructor, you can now simplify the constructor declaration with <>, and the compiler will infer the type.

Issue found using Sonar Cube.


snoopdave:
LGTM. Thanks!

PR #78, opened 2021-03-13
Removed: unused css rule background-repeat from the menu.css file

ymaheshwari1:
We have already defined `no-repeat` in the `background` css property so the `background-repeat` is duplicate so removed that property.


adityasharma7:
Thanks @ymaheshwari1 for your contribution!

PR #80, opened 2021-03-18
Improved: css margin and padding rule to use shorthand property

ymaheshwari1:
Using shorthand property for margin and padding.


adityasharma7:
Thanks @ymaheshwari1 for your contribution!

PR #81, opened 2021-03-19
JSP updates

mbien:
Small JSP related updates. Details in commit msgs.

regarding ea8ec7f
 - MediaFileEdit.jsp textfield size is changed so that the horizontal scroll bar won't appear in the dialog
 - EntrySidebar.jsp text truncation is lowered a little bit so that it doesn't wrap long entry titles to new lines so often


snoopdave:
LGTM

mbien:
thanks for the review

PR #82, opened 2021-03-19
Added Roller 6.1 db table upgrade logic + maven dependency upgrades

mbien:
The table name change requires an upgrade script. Its very generic SQL but don't know on how many DBs this is going to work. I only tested it with HSQLDB.

updated maven dependencies
 - guice 5 can be used again without the noaop tag on latest jdks
 - asm, spring, lucene, velocity etc
 
updated webjars
 - bootstrap 3.4.1 is the last version before 4.x which comes with breaking changes
 
checked JDK16 compatibility
 - required war plugin update in selenium pom

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -39,26 +39,26 @@ limitations under the License.
         <java-mail.version>1.4.7</java-mail.version>
         <jstl.version>1.2</jstl.version>
         <angular.version>1.7.8</angular.version>
-        <ant.version>1.10.8</ant.version>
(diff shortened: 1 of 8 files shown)


snoopdave:
This LGTM except for the `TagDataServlet` warning, which is not something you changed.

mbien:
> This LGTM except for the `TagDataServlet` warning, which is not something you changed.
thanks for reviewing Dave!

regarding the unrelated TagDataServlet warning:
see draft PR #84

PR #83, opened 2021-03-21
Remove provided scope for JSTL

snoopdave:
This PR removes provided scope for JSTL becuase JSTL is not provided by Tomcat. I erroneously added provided scope a couple months ago when I was trying to get the Maven Jetty stuff working.


mbien:
if i remember correctly JSTL libs are supposed to be placed in the classpath of the server/container, not the application. However, these days this should not make any difference anymore since there is usually only one app per server instance - so not much can go wrong.

Keeping it in the roller war for convenience reasons is probably fine - i don't think anyone complained so far :)

PR #84, opened 2021-03-22
Improve input validation

mbien:
Add early input validation for weblog handles and similar parameters so that roller can respond early instead of responding later in the business logic. 

Output encode client provided strings.

Improved Exception handling / Error page mapping.

### DIFF
--- a/app/src/main/java/org/apache/roller/weblogger/webservices/tagdata/TagDataServlet.java
+++ b/app/src/main/java/org/apache/roller/weblogger/webservices/tagdata/TagDataServlet.java
@@ -107,7 +107,7 @@ public class TagDataServlet extends HttpServlet {
         if (request.getParameter("format") != null) {
             
             format = request.getParameter("format");
-            if(!format.equals("json") || !format.equals("xml")) {
+            if(!format.equals("json") && !format.equals("xml")) {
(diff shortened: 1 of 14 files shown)


mbien:
force pushed

* switched from owasp encoder to commons StringEscapeUtils.escape* - no need to add another library for this simple usecase
* validation is currently fairly strict (too strict?) and enforces alphanumeric Strings for blog handles, params and similar tokens

app/src/main/java/org/apache/roller/weblogger/webservices/tagdata/TagDataServlet.java
```diff
@@ -70,36 +72,56 @@ public void doGet(
...
+        }
+        
+        String format = "json";  // default
+        
         if (request.getParameter("format") != null) {
+            
             format = request.getParameter("format");
+            if(!format.equals("json") || !format.equals("xml")) {
+                response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Malformed URL");
+                return;
```
mbien, line 112: todo: this should be && not ||

snoopdave:
Lot's of good changes here. I went through commit-by-commit and did not see any problems.

snoopdave (approved):
LGTM

mbien:
thank you for the review @snoopdave

PR #85, opened 2021-03-26
Fix for failing SubscriptionFunctionalTest

mbien:
PlanetManagerLocalTest leaves a subscription behind which causes SubscriptionFunctionalTest to fail if it is run after it.

This attempts to fix it by removing the subscription in the teardown.


snoopdave (approved):
LGTM

mbien:
thanks dave!

PR #87, opened 2021-05-14
Improved: the UI for the blog entries page

ymaheshwari1:
Made the same UI for the published column and updated column in the blog entries table.
Also, reduced the width to 3% for the edit icon and delete icon column in the same table.

*Before:*

image

*After:*

image


adityasharma7:
LGTM. Thanks @ymaheshwari1

PR #88, opened 2021-05-18
Small performance improvements

mbien:
Noticed that the cache copies its contents twice since flush() + close() are used in sequence in the entire codebase. CachedContent.close() will now only update the content if something changed.

+ added some other minor performance related commits
+ fixed one leaking input stream
+ point release dependency updates


snoopdave (approved):
LGTM!

PR #89, opened 2021-05-21
Improved: tooltip content on the registration form

ymaheshwari1:
1. Improved the tooltip content for the screen name in the registration form

### DIFF
--- a/app/src/main/resources/ApplicationResources.properties
+++ b/app/src/main/resources/ApplicationResources.properties
@@ -1454,7 +1454,7 @@ A-Z and 0-9) and do not use HTML. You will <b>not</b> be able to change this \
 later, but that''s OK because this string will not be publicly displayed.
 
 userRegister.tip.screenName=Your desired screen name (with no HTML). This is \
-is what will be displayed in your weblog and on weblog entries that you write.
+what will be displayed in your weblog and on weblog entries that you write.
 
 userRegister.tip.fullName=Your full name (with no HTML), first and last. Some \
 weblog themes will display this instead of (or in addition to) your screenname.


ymaheshwari1:
Hi,

As I am looking around the roller, found that when using struts2 tags the tooltip attribute does not support HTML, so Should we remove HTML from the tooltip text?

**Text passed as tooltip content:** https://github.com/apache/roller/blob/master/app/src/main/resources/ApplicationResources.properties#L1451**

**UI:**
image

mbien:
the bootstrap ui does still support html in tooltips, its just disabled by default. 
see https://getbootstrap.com/docs/3.4/javascript/#tooltips-options

I am not sure what would be the best way to enable it. Doesn't look like struts lets you set a data attribute.

adityasharma7:
Hi @ymaheshwari1 @mbien,

For the current case, I think we can consider removing the HTML as there are security concerns enabling the HTML mode.

WDYT?

ymaheshwari1:
I think, yes we can do that, as the HTML is used only in some tooltips.

mbien:
> Hi @ymaheshwari1 @mbien,
> 
> For the current case, I think we can consider removing the HTML as there are security concerns enabling the HTML mode.
> 
> WDYT?

yeah sure, there are not that many anyway. I was checking if we could maybe enable some selectively for strings which are known to be static. But yeah lets just remove them.

are those properties also tooltips or are they used somwhere else?
https://github.com/apache/roller/blob/743f1d246deab7d90221cc864093f00c4f672ecd/app/src/main/resources/ApplicationResources.properties#L1474-L1476
since they contain links.

ymaheshwari1:
> OpenID

Hi, These are displayed when the `authMethod != 'LDAP'`. Also, they are not tooltips.

https://github.com/apache/roller/blob/master/app/src/main/webapp/WEB-INF/jsps/core/Register.jsp#L73

But, as they are used in `s:text`, it only considers string in it.
https://struts.apache.org/tag-developers/text-tag.html

mbien:
thanks for the fix looks good.

PR #90, opened 2021-05-31
Parse referrer URL instead of using string value directly.

snoopdave:


mbien:
LGTM, small remark for your consideration:

org.apache.commons.validator.routines.UrlValidator might be better since it does not rely on the exception in the program flow. Lib is already in the dependencies.

example snippet in [doc](https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/UrlValidator.html)

snoopdave:
> LGTM, small remark for your consideration:
> 
> org.apache.commons.validator.routines.UrlValidator might be better since it does not rely on the exception in the program flow. Lib is already in the dependencies.
> 
> example snippet in [doc](https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/UrlValidator.html)

Thanks for the review. I switched over to `URLValidator` as you suggested.

mbien:
@snoopdave excellent. LGTM. (the fixed test looks good too, I was wondering if it was only failing on my machine)

mbien:
4c55a62 regexp eval replacement logic looks good to me

PR #92, opened 2021-06-16
Replaced deprecated rome-fetcher with java 11 HttpClient

mbien:
rome-fetcher can be relatively easy replaced with Java 11 HttpClient while still using the non-deprecated parts of Apache ROME, the only piece missing is the on-disk feed cache.

Interestingly, it appears that the cache was only used as workaround to timestamp feeds which did not have update dates in them. So removing the cache might alter the logic slightly for feed entries without update date since they are always seen as not modified now.

It would be great if someone with Roller Planet experience could check if the aggregator is still working as expected. Tests are green.


snoopdave:
Very nice! don't see any problems here.

snoopdave (approved):
LGTM

PR #64, opened 2020-06-12
fixBug PropertyNotFoundException

DKARAGODIN:
During deploy on clean Wildfly 19 installation there were the foollowing exception:
org.hibernate.PropertyNotFoundException: Could not locate setter method for property [org.apache.roller.weblogger.pojos.WeblogEntryComment#emptyString]
Adding missing <transient> element for emptyString property in orm descriptor fixes the problem.

PR #93, opened 2021-06-27
fix for PropertyNotFoundException when started in WildFly

mbien:
second attempt of https://github.com/apache/roller/pull/64

removed getEmptyString() since listValue="" has the same effect and makes WildFly happy too.


mbien:
integrating since trivial + #64 was already reviewed

PR #95, opened 2021-07-08
Updated: the main menu to have proper space when displaying permission message

ymaheshwari1:
Added space between the permission message.

Before:
image

After:
image


mbien:
good catch! merging.

PR #96, opened 2021-07-10
Deprections, compiler warning fixes, local refactorings and JDK 17 support

mbien:
Worked myself though the compiler warnings. This looks like a lot of changes, but the individual changes are all very local.

* since most of the rawtype warnings were on Collections, I updated a lot of the code to use more modern APIs instead of only fixing the warnings (e.g List.of, streams, ...)
* extracted common reflection code to a utility class
* got rid of some deprecations
* some cleanup and improvements (~400 lines less code)
* JDK 17 support (runs still on 11)


snoopdave (approved):
Works for me. I built and ran on JDK17 and did not notice any problems.

I did notice one error in tests but I think it  is fixed in master:

```
[ERROR] Errors:
[ERROR]   SingleThreadedFeedUpdaterTest.testUpdateSubscription:72 » Updater Error fetchi...
```

mbien:
> Works for me. I built and ran on JDK17 and did not notice any problems.
> 
> I did notice one error in tests but I think it is fixed in master:
> 
> ```
> [ERROR] Errors:
> [ERROR]   SingleThreadedFeedUpdaterTest.testUpdateSubscription:72 » Updater Error fetchi...
> ```

hmm. This test usually fails when rollerblogger.org is temporary unreachable since it relies on its feed.

I tested this PR again and everything was green except the IPBanListTest, but the test itself has some file-io race conditions on my system - but this would be unrelated to this PR. Going to integrate this but keep an eye on the test :)

PR #97, opened 2021-07-11
EncodePreTagsPlugin

mbien:
Simple weblog entry plugin which can be used to escape angle brackets from text within pre tags.

It understands the common pattern of pre and code tag in succession which is used by javascript source code highlighters and doesn't escape code tags directly following pre.

Hopefully useful for programmers which like to blog in html mode and simply paste code into pre tags without breaking the page.


snoopdave (approved):
Very nice!

mbien:
just found a bug while writing a blog entry :) I always forget that a Matcher doesn't like '$'' signs in the input by default. Have to fix this first - should be easy.

snoopdave (approved):
LGTM

mbien:
> LGTM
you are too fast for me Dave, now its fixed :)

PR #99, opened 2021-07-18
Updated: the header to show the tagline only when its available

ymaheshwari1:
Added an if condition to check whether the weblog has a tagline defined or not and if there is a tagline present then only displaying the tagline.

Before:
Tagline is always displayed.
image

After:
Tagline is only displayed when its defined
image

If tagline is not present then nothing is displayed. 
image


mbien:
Hi Yash, have you checked if this can also happen with the other themes?

ymaheshwari1:
Hi @mbien I am checking other themes for the same.

mbien:
lgtm, integrating

PR #103, opened 2021-09-13
"/tags/" appears twice in the cache key.

mbien:
Small fix. Did not impact functionality, just the way cache keys were generated.

bug introduced in  7d07a84


snoopdave (approved):
AOK

PR #104, opened 2021-09-14
Inefficient Usage of Java Collections

FastAtlas:
Hi,

We find that there are several inefficient usages of Java Collections:

1. The contains method is invoked upon a list object. We recommend replacing it with a HashSet.
2. There is no iteration occurring upon a TreeMap, thus the insertion order does not matter. We recommend replacing it with a HashMap.

We discovered the above inefficient usage of containers by our tool Ditto. The patch is submitted. Could you please check and accept it? We have tested the patch on our PC. The patched program works well.

Bests

Ditto

### DIFF
--- a/app/src/main/java/org/apache/roller/weblogger/business/jpa/JPAWeblogManagerImpl.java
+++ b/app/src/main/java/org/apache/roller/weblogger/business/jpa/JPAWeblogManagerImpl.java
@@ -37,7 +37,6 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.TreeMap;
 
 import org.apache.roller.weblogger.business.MediaFileManager;
 import org.apache.roller.weblogger.business.UserManager;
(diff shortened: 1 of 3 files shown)


mbien:
hello @DittoTool

i took a quick look, and beside it not compiling:
* 850c7a5 getWeblogHandleLetterMap() can't be changed since it is a public service and returns an alphabetically sorted index
* 7e46a11 RollerPermission would need more work, but since roller has very few permissions, it wouldn't really make a difference performance wise. Additionally: the permissions are persisted as comma separated String, so changing the order randomly might lead to other problems and would need more investigation and testing.
* b24e89c is example code which uses a hardcoded list of 3 items.

please don't claim that you tested a PR when the changes don't compile (i will give you the benefit of the doubt and assume that the commit is missing something).

(i can't find dittotool on the apache contributors list https://people.apache.org/committer-index.html, so you might have to sign the CLA first before contributing to an apache project: https://www.apache.org/licenses/contributor-agreements.html )

PR #111, opened 2021-12-22
bugfix: SearchServlet forgot to copy the user to WeblogPageRequest.

mbien:
Template model of the search results page always thought the user was not logged in.

fix is here f1f80002ace0c5983bb4f7437110b4216e62cba9 other commit tries to improve readability a little bit.


mbien:
@gregh3269 this should fix the issue you mentioned on the dev list.

snoopdave (approved):
Nice work.

PR #112, opened 2021-12-25
Minor fixes in Velocity Templates and dependency updates

mbien:
- Apache Rome had a release (yey)
 - another week, another CVE https://logging.apache.org/log4j/2.x/
 - templates should use `$foreach.index` in loops, old `$velocityCount` isn't working anymore
 - another stream was left open
 - Registered some maven actions for NetBeans


snoopdave (approved):
Velocity changes LGTM. Do the Netbeans changes allow you to debug tests in the IDE?

mbien:
> Velocity changes LGTM. Do the Netbeans changes allow you to debug tests in the IDE?

Its mostly for registering non-standard maven goals. like jetty:run, or the selenium tests which have to specify a path to the "geckodriver". Pressing the run/debug button on Roller webapp or the Selenium tests will do the right thing. Debugging junit did actually work before already since netbeans knows how to use that plugin.

There is also a "display-dependency-updates" action registered for the root project to quickly check if something can be updated.

I am not feeling strongly about it. If you don't want it in the repository I can leave it out and put nbactions.xml on gitignore.

snoopdave:
Cool. I'm all for it if it makes things easier for folks using Netbeans.

PR #113, opened 2021-12-25
Added github actions for basic build/test setup.

mbien:
- builds/tests on JDK 11 and 17
- uploads junit report
- disabled a failing test for now

PR #114, opened 2021-12-28
Merged CodeQL workflow into main workflow.

mbien:
followup on https://github.com/apache/roller/pull/113

This is a workaround which ensures that the test reports end up in the correct workflow.

example:
JUnit reports of:
https://github.com/apache/roller/actions/runs/1627718913
ended up here:
https://github.com/apache/roller/actions/runs/1627718912

If there is just one workflow, they can't end up in the wrong one :)


mbien:
looks like the CodeQL alerts are going to be reset (reported as new issues, while the old are removed) since the job is started from a different file.
but the aggregation is working: https://github.com/apache/roller/pull/114/checks

mbien:
closing in favor of #119

PR #115, opened 2022-04-01
dependency updates, added bouncycastle, minor log level update.

mbien:
- **notable dependency update: spring 5.3.18** with **security** fixes for "[spring4shell](https://security.snyk.io/vuln/SNYK-JAVA-ORGSPRINGFRAMEWORK-2436751)"
 - bouncy castle was added so that the popular Argon2 pw encoder can be used

available (non legacy) encoders: bcrypt, pbkdf2, scrypt, argon2

**Reminder how to change the encoder:**
`passwds.encryption.algorithm=argon2`, default remains `pbkdf2`
automatic migration happens on next pw change

**If one of the (very, very) old legacy codecs is still in use, roller will need a hint for migration**
`passwds.encryption.lazyUpgradeFrom=`
possible options: `plaintext, MD5, SHA-1` can be left empty if the db doesn't contain any legacy encoded pws

@snoopdave might be a good time for another point release :)

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -41,22 +41,23 @@ limitations under the License.
         <angular.version>1.7.8</angular.version>
         <ant.version>1.10.12</ant.version>
         <asm.version>9.2</asm.version>
+        <bouncycastle.version>1.70</bouncycastle.version>
         <commons-validator.version>1.7</commons-validator.version>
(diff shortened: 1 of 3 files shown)


mbien:
merging (without review) to make it easier for users to build roller with the spring fixes.

PR #116, opened 2022-04-02
upload dev build + add JDK 18 to the matrix + version bumps.

mbien:
upload a dev build as CI step.


mbien:
verified that it works. It will only upload the artifact during integration, not during PR checks to conserve resources.
example at bottom of the page:
https://github.com/apache/roller/actions/runs/2080694025

link to all builds for master branch:
https://github.com/apache/roller/actions/workflows/main.yml?query=branch%3Amaster

retention is set to 90 days

PR #119, opened 2022-11-14
minor dependency updates + test-summary workflow action

mbien:
- the test-summary action has [been allowed by infra](https://issues.apache.org/jira/browse/INFRA-23445) for CI use and works fairly well for the NetBeans project where we use it since then. The only annoyance is that it ignores errors, while it will pick up junit failures, but a PR is pending for that.
It writes test reports directly to the workflow summary page (not to checks, which felt like a workaround anyway).
- test runners should have geckodriver preinstalled, lets see if we can simply set a path to it.
- enabled xserver for headful integration testing
- minor maven dependency updates
- gh action version updates since they cause warnings
- bump JDK 18 to ~19~ 20 in the test matrix

tested locally.. lets see if everything builds with CI too

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -39,24 +39,24 @@ limitations under the License.
         <java-mail.version>1.4.7</java-mail.version>
         <jstl.version>1.2</jstl.version>
         <angular.version>1.7.8</angular.version>
-        <ant.version>1.10.12</ant.version>
-        <asm.version>9.3</asm.version>
+        <ant.version>1.10.13</ant.version>
+        <asm.version>9.4</asm.version>
         <bouncycastle.version>1.70</bouncycastle.version>
         <commons-validator.version>1.7</commons-validator.version>
         <commons-beanutils.version>1.9.4</commons-beanutils.version>
         <commons-codec.version>1.15</commons-codec.version>
-        <eclipse-link.version>3.0.2</eclipse-link.version>
+        <eclipse-link.version>4.0.1</eclipse-link.version>
         <guice.version>5.1.0</guice.version>
-        <log4j2.version>2.17.2</log4j2.version>
-        <lucene.version>9.2.0</lucene.version>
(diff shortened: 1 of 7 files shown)


mbien:
ok thats weird. I can't explain right now why this PR started failing. Can't reproduce this issue locally and I basically reverted everything already.

mbien:
there you go. Some test annotations were missing.

Whats interesting however is that the tests run in a different order on github than they do on my system locally. Once the MediaFile test started running first, it failed since it didn't boostrap roller.

mbien:
some notes on dependency bumps:
 - Spring 6 is released but it requires JDK 17, so I kept it at 5.x for now since Roller is using 11 still
 - Struts [2.5.30](https://cwiki.apache.org/confluence/display/WW/Version+Notes+2.5.30) and later has some breaking changes, so I kept it at 2.5.29 for now.

mbien:
- rebased on top of latest master
- refreshed minor dep updates
- switched from JDK 19 to 20 since a release happened in the meantime

tested locally

app/pom.xml
```diff
@@ -39,24 +39,24 @@ limitations under the License.
...
-        <lucene.version>9.2.0</lucene.version>
+        <log4j2.version>2.20.0</log4j2.version>
+        <lucene.version>9.5.0</lucene.version>
         <oauth-core.version>20100527</oauth-core.version>
-        <maven-war.version>3.2.3</maven-war.version>
+        <maven-war.version>3.3.2</maven-war.version>
         <maven-surefire.version>2.22.2</maven-surefire.version>
         <maven-antrun.version>1.0b3</maven-antrun.version>
-        <rome.version>1.18.0</rome.version>
+        <rome.version>1.19.0</rome.version> <!-- locked in place since next version removes popono -->
```
mbien, line 56: another lib which is now locked since rome popono is gone

mbien:
all green. @snoopdave approved via dev list -> merging

PR #121, opened 2023-03-07
Start Derby from Java so tests can run without Maven

snoopdave:
This PR removes the Maven Plugin that was used to start/stop Derby and re-introduces some old code that starts Derby via Java and creates the Roller schema before tests run and stops Derby after they complete.

Also, changes version to `6.1.2-SNAPSHOT`


mbien:
the integration tests don't seem to like this change unfortunately

snoopdave:
> the integration tests don't seem to like this change unfortunately

Yep. Thanks. I'm working on a fix for this.

PR #127, opened 2023-06-22
Some field validation CSP updates

mbien:
The content security policy might break some blogs if they are used to embed content from other domains. However, the templates are all customizable so there is a way around it for the admins.


snoopdave:
Dang! I did not see this PR

snoopdave (approved):
LGTM!

PR #128, opened 2023-07-01
Santize some inputs and clarify docs

snoopdave:
This PR sanitizes a couple of inputs and adds some docs clarifications.


mbien (approved):
looks good. much better than in the bean.

snoopdave:
@mbien thanks for the review

PR #131, opened 2023-09-17
Add a Jetty event listener to start Derby when Roller is run via mvn jetty:run

snoopdave:
This PR fixes a regression caused by switching Roller from using the Maven `btmatthews` `inmemdb` plugin and instead starting Derby by calling the embedded Derby server API directly (which works when tests are run either in IntelliJ or in Maven).  The problem was that I forgot to add the same logic to the Maven Jetty plugin so that Derby is started when Roller is run via `mvn jetty:run`.

The Jira ticket: https://issues.apache.org/jira/browse/ROL-2182

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -36,7 +36,6 @@ limitations under the License.
         <servlet.version>4.0.1</servlet.version>
         <jsp-api.version>2.2</jsp-api.version>
         <java-activation.version>1.2.0</java-activation.version>
-        <java-mail.version>1.4.7</java-mail.version>
         <jstl.version>1.2</jstl.version>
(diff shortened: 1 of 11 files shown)


mbien:
unfortunately this breaks the integration tests :(

snoopdave:
Bummer. I'll take another look over the weekend. I thought I had fixed the IT tests, except for the jdk21 run.

mbien:
@snoopdave oh you are right - I looked at the wrong workflow run! All good, my mistake.

the jdk 21 run is easy to fix. its just the setup action requesting an early access build - i can fix this later.

PR #134, opened 2024-02-03
Validate tz

snoopdave:
Add some input validations and prepare for next release which will be 6.1.3.


app/src/main/java/org/apache/roller/weblogger/pojos/WeblogBookmarkFolder.java
```diff
@@ -188,4 +189,9 @@ public List<WeblogBookmark> retrieveBookmarks() throws WebloggerException {
         return bmgr.getBookmarks(this);
     }
 
+    public void sanitize() {
+        // Conditionally sanitize fields not validated by Struts Validator
+        setName(HTMLSanitizer.conditionallySanitize(getName()));
+    }
```
mbien, line 195: question: can't this be done in the setters? This would ensure that the fields are always clean without having to call the method. Or is this due to serialization or something like that?

snoopdave: I think you're right and that is a better approach. I will adjust.

mbien (approved):
looks good!

mbien (commented):
actually: the cleanup would need to happen in the constructor too I believe, since this would sideline the setters otherwise.

(constructor could call the setter potentially)

PR #137, opened 2024-09-04
Deployment Automation

Chris9419:
### Summary
This pull request adds the following deployment scripts and configurations:

- `Vagrantfile` for setting up the development environment
- `Dockerfile` for containerizing the application
- `docker-compose.yml` for orchestrating multi-container setups
- `deploy.sh` script for automating the deployment process

### Why This Change Is Necessary
These changes are necessary to automate and streamline the deployment process, making it reproducible and easy to manage.

### Testing Done
- Verified that the deployment process works as expected by following the documented steps.
- Tested the Docker containers to ensure the application runs properly with PostgreSQL.

### Checklist
- [x] Code adheres to the project's style guidelines
- [x] Changes have been tested
- [x] Relevant documentation has been updated


.vagrant/machines/default/virtualbox/action_provision
```diff
@@ -0,0 +1 @@
+1.5:a1cf4c44-6035-470e-b080-fd4088b2dc79
```
snoopdave, line 1: I don't understand why any of the changes under `.vagrant` are necessary in this PR.

.vagrant/machines/default/virtualbox/private_key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
snoopdave, line 1: Seems like a bad practice to include private keys in our repo.

.vagrant/rgloader/loader.rb
```diff
@@ -0,0 +1,9 @@
+# This file loads the proper rgloader/loader.rb file that comes packaged
```
snoopdave, line 1: Why is anything related to Vagrant in this PR, we've got Docker and docker-compose already why do we need Vagrant?

Dockerfile
```diff
@@ -21,65 +21,65 @@
...
-# Add Roller entry-point and go!
-
-COPY --from=builder /project/docker/entry-point.sh /usr/local/tomcat/bin
-COPY --from=builder /project/docker/wait-for-it.sh /usr/local/tomcat/bin
-RUN chgrp -R 0 /usr/local/tomcat
-RUN chmod -R g+rw /usr/local/tomcat
-
-WORKDIR /usr/local/tomcat
-CMD [ "/usr/local/tomcat/bin/entry-point.sh" ]
+    FROM maven:3-openjdk-17-slim as builder
```
snoopdave, line 24: Formatting changes should not be included in PRs, only changes essential to the PR.

Vagrantfile
```diff
@@ -0,0 +1,99 @@
+# -*- mode: ruby -*-
```
snoopdave, line 1: again, why do we need Vagrant when we have Docker and Docker compose?

docker-compose.yml
```diff
@@ -1,47 +1,25 @@
-
```
snoopdave, line 1: Why does this PR remove the license header? Also, any new files shiould include the license header.

docker-compose.yml
```diff
@@ -1,47 +1,25 @@
...
-           source: ./docker/roller-data
-           target: /var/lib/roller/data
-
+  postgresql:
+    image: "postgres:10.0"
+    ports:
+      - "5432:5432"
+    volumes:
+      - type: bind
+        source: C:/roller/docker/postgresql-data
```
snoopdave, line 9: Is there no way to specify a Docker file like this without OS specific paths like this?

snoopdave (commented):
I don't understand the point of this PR as we already have the ability to produce a Docker file and we provide an example docker-compose.yml setup. Also, this PR introduces arbiutrary formatting changes that are not needed and make it difficult to see what has actally changed.

.vagrant/machines/default/virtualbox/private_key
mbien: agreed, private keys should not be in repos many reasons.

mbien (commented):
unless there is a committer who is willing to look over this and maintain the vagrant config, i am going to -1 this.

The main artifacts of this repo are the zip/war. Everything what comes after this can be customized by the user who deploys roller. There is also a docker file serving as example deployment / getting started script.

You could of course host a sample vagrant config in your own repo, but I don't see the purpose of this in the main repo right now esp since we would have to maintain this forever.

repos should also not contain private keys, system specific paths etc but dave already annotated those bits.

snoopdave:
I don't think this is work that we want to merge into Roller.

PR #145, opened 2024-12-20
Fix CVE-2024-53677

loganaden:


snoopdave (changes requested):
With this change 1) Roller starts but returns 500 error on request and 2) The Selenium tests fail. I may be wrong about this, but I think migrating to Struts 6 is going to be take a bit more effort.

app/pom.xml
```diff
@@ -58,7 +58,7 @@ limitations under the License.
         <slf4j.version>2.0.16</slf4j.version>
         <spring.version>5.3.39</spring.version>
         <spring.security.version>5.8.14</spring.security.version>
-        <struts.version>2.5.29</struts.version> <!-- .30+ breaks selenium tests -->
+        <struts.version>6.4.0</struts.version> <!-- .30+ breaks selenium tests -->
```
mbien, line 61: the comment next to it means that any update beyond `.30` will require reading the struts migration guide and resolve problems since the very next release has breaking changes in some areas.

PR #149, opened 2025-02-01
Dependency updates

mbien:
- only point releases
 - all tests pass on JDK 24
 - ~unit tests would pass on JDK 24 if bytebuddy would be bumped (not done since transitive mockito dependency, this should resolve itself automatically)~
 - added JDK 23 to test matrix

haven't deployed this version yet. leaving it as draft till then

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ '11', '17', '21' ]
+        java: [ '11', '17', '21', '23' ]
 
     steps:
       - name: Set up JDK ${{ matrix.java }} 
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -38,28 +38,28 @@ limitations under the License.
         <java-activation.version>1.2.0</java-activation.version>
         <jstl.version>1.2</jstl.version>
         <angular.version>1.7.8</angular.version>
-        <ant.version>1.10.14</ant.version>
-        <asm.version>9.7</asm.version>
+        <ant.version>1.10.15</ant.version>
+        <asm.version>9.7.1</asm.version>
         <bouncycastle.version>1.70</bouncycastle.version>
         <commons-validator.version>1.9.0</commons-validator.version>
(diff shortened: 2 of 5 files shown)


snoopdave:
Thanks @mbien! This looks good. I don't know what you mean about the Mokito thin: "this should resolve itself automatically"?

mbien:
> I don't know what you mean about the Mokito thing: "this should resolve itself automatically"?

Hi @snoopdave! At some point mockito will release a new version which will bump its dependencies to JDK 24 compatible versions. At that point it will "just work" after roller updates the mockito version without having to declare a direct (test) dependency on bytebuddy. That is what I meant by this will fix itself after a while.

(I manually tested bumping bytebuddy to the latest version and tests passed on JDK 24)

mbien:
```
mvn eu.maveniverse.maven.plugins:toolbox:gav-tree -Dgav=org.mockito:mockito-core:5.14.1
[INFO] --- toolbox:0.6.1:gav-tree (default-cli) @ standalone-pom ---
[INFO] org.mockito:mockito-core:jar:5.14.1
[INFO] ├─net.bytebuddy:byte-buddy:jar:1.15.3 [compile]
[INFO] ├─net.bytebuddy:byte-buddy-agent:jar:1.15.3 [compile]
[INFO] ╰─org.objenesis:objenesis:jar:3.3 [runtime]
```

btw https://github.com/raphw/byte-buddy/releases/tag/byte-buddy-1.16.0

> Include support for JDK Class File API.

bytebuddy 1.16 and later should no longer cause this kind of problem thanks for using plain JDK APIs. So mockito won't have to update the dependency that often.

mbien:
@snoopdave oops i didn't realize that there is already a newer mockito available - will update this PR

mbien:
updated, everything passes on JDK 24 now

snoopdave (approved):
LGTM and checks are passing 🐝

snoopdave:
@mbien are we good to merge this? I'd like to get it into 6.1.5 rc1

mbien:
@snoopdave Should be fine - although I haven't deployed it this time around for manually testing. JQuery update is always a bit risky since it probably doesn't have any test coverage. Rest should be good.

PR #150, opened 2026-01-23
build and sonar added

van-123-egg:


mbien:
I don't think we need sonar atm - also couldn't find a discussion about this on the mailing lists, closing

PR #151, opened 2026-02-24
Fix incorrect PostgreSQL port in Dockerfile

MustafaCelal:
Description:
This PR corrects a port mismatch in the Dockerfile that prevents the Apache Roller auto-installer from connecting to the database during the bootstrap process.

Changes:

Updated DATABASE_HOST from postgresql:5434 to postgresql:5432 to match the default PostgreSQL port and the configuration in docker-compose.yml.

Reasoning:
The previous configuration pointed to port 5434, while the docker-compose.yml and standard PostgreSQL images use port 5432. This discrepancy leads to an "Unknown Error" because the application cannot determine the database status or bootstrap itself.


mbien (approved):
makes sense to me. Not sure if it was intended or not but it was set to 5434 since the initial commit.

PR #153, opened 2026-07-28
Update install guide with working JavaMail download links

mraible:
## Summary

- Replaced the dead JavaMail download link (`https://java.net/projects/javamail/pages/Home`) with direct Maven Central links for `jakarta.mail-1.6.7.jar` and `jakarta.activation-1.2.2.jar`
- Added an explanation for *why* the JavaMail JARs are required even without email features: Roller uses `javax.mail.internet.InternetAddress` for email address validation in core code (`Utilities.java`), so the dependency is loaded at runtime regardless of mail configuration
- Updated the SMTP Javadoc link from `javamail.java.net` to the current Eclipse project URL

## Context

Without the JavaMail JARs in Tomcat's `lib/`, Roller throws a 500 error on login with `ClassNotFoundException: javax.mail.internet.AddressException`. The install guide already says to add these JARs, but the download link has been dead since java.net was shut down, making it easy to skip or miss this step.

### DIFF
--- a/docs/roller-install-guide.adoc
+++ b/docs/roller-install-guide.adoc
@@ -479,7 +479,7 @@ password in 'mail/Session' Resource.
 ----
 
 The Java mail properties listed above are defined here:
-https://javamail.java.net/nonav/docs/api/com/sun/mail/smtp/package-summary.html.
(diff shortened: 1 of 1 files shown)


snoopdave (approved):
LGTM

PR #154, opened 2026-08-11
ROL-2183: Migrate from javax to Jakarta EE 11

mraible:
Migrates Apache Roller from Java EE 8 (`javax.*`) to Jakarta EE 11 (`jakarta.*`), upgrading all major frameworks. The target started as EE 10; Copilot review pointed out Spring Framework 7 is only supported on the EE 11 baseline (Servlet 6.1), and the containers all support it, so the servlet API, Jetty, and Tomcat moved up. The branch name keeps its original ee-10 form. Each commit corresponds to a migration phase so reviewers can follow the progression. JIRA: [ROL-2183](https://issues.apache.org/jira/browse/ROL-2183)

**Target versions:** Java 17, Servlet 6.1, Struts 7.1.1, Spring 7.0.8, Spring Security 7.0.6, EclipseLink 5.0.1, Jetty 12.1.12, Tomcat 11

**Removed functionality** — called out explicitly so nothing merges silently:

- OAuth 1.0a is gone: the AtomPub API's `oauth` authentication option and the consumer key management UI. The `net.oauth` library is unmaintained and javax-only, and the protocol was obsoleted by OAuth 2.0 (RFC 6749, 2012). AtomPub authentication now offers `basic` and `wsse` only. The OIDC PR stacked on this one adds OAuth 2.0 for browser login; it does not cover AtomPub.
- OpenID 2.0 login no longer works at this level: `spring-security-openid` was deleted upstream in Spring Security 6. The stacked OIDC PR replaces it with OpenID Connect and retires the `openid`/`db-openid` auth method values.

Everything else API-facing survives: Blogger/MetaWeblog XML-RPC and AtomPub are preserved by forking their javax-only servlets into Roller (`xmlrpc-server` and `rome-propono` have no Jakarta releases), and WebJar URLs are unchanged (Servlet 3.0+ serves `META-INF/resources` natively, so the javax-only webjars servlet could simply be dropped).

**Behavior notes for reviewers:**

- The forked Atom adapter keeps propono’s null path info normalization, so a request to the exact `/roller-services/app` mapping serves the service document (caught by Copilot review; guarded by a smoke test in the Playwright PR).
- docker-compose binds a new `postgresql-16-data` directory because PostgreSQL 16 cannot start on a data directory from the PostgreSQL 10 image the old file used (also from Copilot review).
- AtomPub basic authentication is fixed: it verified the password against a field that is never set at that point, so every request got a 401. The bug exists on master too; the new web services smoke test in this stack caught it.
- Media upload actions now implement Struts 7's `UploadedFilesAware` — Struts 7 removed the old `File[]`/setter injection, which silently broke uploads until Greg Huber caught it on dev@roller. Fixed and covered by a browser test in the Playwright PR of this stack.
- Form markup at this level is emitted in Bootstrap 5 flavor by struts2-bootstrap-plugin 6.1.0 (the Jakarta-native release, which also supplies the form theme templates Roller relies on) while the page CSS is still Bootstrap 3, so a few forms look rough here (stacked buttons, missing tooltip icons). The Bootstrap 5 PR stacked on this one reconciles them.

**Phase 0** — Java 17 baseline, Derby 10.16.1.1 (old version does not support Java 17), SLF4J 2.x bridge fix

**Phase 1** — Remove unmaintained/obsolete dependencies while still on javax (testable incrementally): OAuth 1.0a (14 files deleted, 17 edited), spring-security-openid (filter, beans, OpenID URL detection in RollerUserDetailsService/RollerSession), XmlRpcServlet and AtomServlet forked in, struts2-bootstrap-plugin upgraded 4.0.0 → 6.1.0.

**Phase 2** — The big switch (one coordinated change since Struts 7, Spring 7, and Jakarta APIs are tightly coupled):
- 320 `javax.servlet`/`javax.mail`/`javax.activation` imports converted to `jakarta.*` across 89 Java files
- `com.opensymphony.xwork2.*` imports migrated to `org.apache.struts2.*`
- Deprecated Struts interfaces (ServletRequestAware, RequestAware, HttpParametersAware) replaced with ActionContext lookups
- Struts 7 runtime fixes: parameter binding is annotation-gated by default (`struts.parameters.requireAnnotations=false` since Roller's actions predate `@StrutsParameter`), an OGNL allowlist for Roller packages, the static content path, and the prepare interceptor re-enabled ahead of params
- Spring Security AccessDecisionManager/RoleVoter removed, migrated to expression-based access control; Spring Security 7's authorization manager denies unmatched requests, so a `permitAll` catch-all preserves the old open-by-default behavior for public URLs
- All Maven dependencies swapped to Jakarta equivalents
- persistence.xml + 29 ORM files updated to Jakarta JPA namespace
- web.xml, struts.xml, tiles.xml, security.xml config updated
- JSP error pages updated (`javax.servlet.error.*` → `jakarta.servlet.error.*`)
- Password encoder constructors updated for Spring Security 7 API changes

**Phase 3** — Build and deployment tooling:
- Jetty Maven plugin upgraded from 10 to 12 (ee10 module for Jakarta servlet support)
- `mvn jetty:run` fixed for the Jetty 12 plugin (`jettyXml` was renamed `jettyXmls` and silently ignored, Jetty 12 jetty.xml syntax, a mail session built from angus-mail, commons-dbcp2) and it now comes up with a ready-to-use in-memory Derby schema
- Dockerfile updated: Tomcat 9 → 10.1, javax.mail → angus-mail, PostgreSQL driver updated
- docker-compose.yml updated: PostgreSQL 10 → 16
- CI matrix is now JDK 17, 21, and 25 (Java 11 dropped for the 17 baseline; 23 replaced by the 25 LTS, which required Mockito 5.23.0 to instrument JDK 25 class files)

All 157 unit tests pass on JDK 17, 21, and 25.


mraible:
The CI failures here predate the latest push: both runs failed the same way, with the it-selenium test meeting a 503 because the webapp never actually deployed. Debugging that locally turned up a chain of real migration bugs that the branch's only container test had been hiding, now fixed in the last four commits:

- The Jetty 12 plugin renamed `jettyXml` to `jettyXmls` and silently ignored the old name, so the it-selenium harness never loaded jetty.xml, never bound the JNDI datasource, and never started the Derby network server (`Nothing to bind for name jdbc/rollerdb` in the earlier run's log).
- webjars-servlet-2.x extends the javax `HttpServlet`, so on a Jakarta container it threw `NoClassDefFoundError` at startup and took the whole context down. WebJar assets are now served natively from `META-INF/resources` at the same `/webjars` paths.
- Spring Security 7's XML config authorizes through `AuthorizationManager`, which denies requests that match no `intercept-url` rule; the old `FilterSecurityInterceptor` allowed them. Every unlisted URL, including the login page itself, redirected to the login page in a loop. An explicit `/**` permitAll rule restores the old behavior.
- struts2-bootstrap-plugin had been removed as an unused taglib, but it also supplies the FreeMarker templates behind the `theme="bootstrap"` attribute that nearly every form JSP uses, so any page with a form failed with `TemplateNotFoundException`. The Jakarta-native 6.1.0 restores the theme; it emits Bootstrap 5 markup against the current Bootstrap 3 CSS, so some form styling is off until a follow-up converts the UI to Bootstrap 5.

The full Selenium journey (register, log in, create a weblog, publish an entry) and the 157 unit tests pass locally on this branch after these changes.

app/pom.xml
```diff
@@ -46,21 +45,20 @@ limitations under the License.
...
+<maven-war.version>3.4.0</maven-war.version>
         <maven-surefire.version>3.5.2</maven-surefire.version>
         <maven-antrun.version>1.0b3</maven-antrun.version>
         <rome.version>1.19.0</rome.version> <!-- locked in place since next version removes popono -->
         <slf4j.version>2.0.16</slf4j.version>
-        <spring.version>5.3.39</spring.version>
-        <spring.security.version>5.8.14</spring.security.version>
-        <struts.version>2.5.29</struts.version> <!-- .30+ breaks selenium tests -->
+        <spring.version>7.0.8</spring.version>
+        <spring.security.version>7.0.6</spring.security.version>
```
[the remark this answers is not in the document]
mraible: Good catch. The PR now targets EE 11: Servlet API 6.1.0, the ee11 flavor of the Jetty plugin at 12.1.12, and Tomcat 11 in the Docker image (JSP was already at 4.0). All 157 unit tests and the full browser suite from the stacked Playwright PR pass on Jetty 12.1. The PR title and description are updated accordingly.

app/src/main/java/org/apache/roller/weblogger/webservices/atomprotocol/RollerAtomRequestImpl.java
```diff
@@ -0,0 +1,135 @@
...
+    private final HttpServletRequest wrapped;
+
+    public RollerAtomRequestImpl(HttpServletRequest request) {
+        this.wrapped = request;
+    }
+
+    @Override
+    public String getPathInfo() {
+        return wrapped.getPathInfo();
+    }
```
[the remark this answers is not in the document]
mraible: Confirmed against the original: rome-propono's AtomRequestImpl normalizes a null path info to an empty string, and the fork dropped that. Restored, and the web services smoke test in the stacked Playwright PR now requests the exact /roller-services/app mapping so this stays covered.

docker-compose.yml
```diff
@@ -19,12 +19,10 @@
 # Example Docker Compose setup for running Roller and PostgreSQL locally
 
 
-version: '3.7'
-
 services:
 
    postgresql:
-      image: "postgres:10.0"
+      image: "postgres:16"
```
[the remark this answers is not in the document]
mraible: docker-compose now binds ./docker/postgresql-16-data with a comment explaining that PostgreSQL 16 cannot start on a data directory from the old PostgreSQL 10 image, so existing checkouts keep their old data untouched instead of failing on it.

app/src/main/resources/org/apache/roller/weblogger/pojos/WeblogTemplate.orm.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<entity-mappings version="2.0" xmlns="http://java.sun.com/xml/ns/persistence/orm"
+<entity-mappings version="3.1" xmlns="https://jakarta.ee/xml/ns/persistence/orm"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
-    xsi:schemaLocation="http://java.sun.com/xml/ns/persistence/orm http://java.sun.com/xml/ns/persistence/orm_2_0.xsd">
+    xsi:schemaLocation="https://jakarta.ee/xml/ns/persistence/orm https://jakarta.ee/xml/ns/persistence/orm/orm_3_1.xsd">
```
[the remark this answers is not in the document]
mraible: Checked against jakarta.ee: it is the other way around. https://jakarta.ee/xml/ns/persistence/orm/orm_3_1.xsd (the URL these files use) returns HTTP 200 and serves the Jakarta Persistence ORM 3.1 schema, while https://jakarta.ee/xml/ns/persistence/orm_3_1.xsd (the suggested correction) returns 404. The mappings keep the working URL. This applies to all 29 of these comments, resolving them together.

it-selenium/logs/roller-2026-08-11-165858.log.gz
snoopdave, line 1: Please remove all of these logs from the PR

app/src/main/resources/struts.xml
```diff
@@ -17,10 +17,26 @@
...
+    <!-- Struts serves its bundled JS/CSS (tooltips, etc.) from this path, which
+         must match the /struts/* filter-mapping in web.xml. Struts defaults it
+         to /static, which Roller does not map. -->
+    <constant name="struts.ui.staticContentPath" value="/struts"/>
+
+    <!-- Struts 7 only binds request parameters to properties annotated with
+         @StrutsParameter. Roller's actions predate that annotation, so without
+         this no form in the application can submit data. Annotating the action
+         beans would be the stricter alternative. -->
+    <constant name="struts.parameters.requireAnnotations" value="false"/>
```
snoopdave, line 38: AI review flagged this change. 

The alternative to disabling this is adding annotations to each form input field? I wonder 1. how much of a security risk this is and 2. how much work we would have to do to add annotations to every input field.

app/src/main/webapp/WEB-INF/web.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<web-app xmlns="http://xmlns.jcp.org/xml/ns/javaee"
+<web-app xmlns="https://jakarta.ee/xml/ns/jakartaee"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/web-app_4_0.xsd"
-         version="4.0">
+         xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-app_6_0.xsd"
+         version="6.0">
```
snoopdave, line 5: Should this be 6.1 instead of 6.0?

.gitignore
```diff
@@ -28,3 +28,4 @@ docker/postgresql-data
 docker/roller-data
 assembly-release/release.sh
 it-selenium/overlays/
+logs/
```
snoopdave, line 31: do we also need to ignore `docker/postgresql-16-data`?

app/src/main/java/org/apache/roller/weblogger/ui/struts2/util/UIAction.java
```diff
@@ -88,13 +87,14 @@ public abstract class UIAction extends ActionSupport
...
     }
-	
-    @Override
-	public void setRequest(Map<String, Object> map) {
-		this.salt = (String) map.get("salt");
-	}
 
 	public String getSalt() {
+		if (salt == null) {
+			HttpServletRequest request = ActionContext.getContext().getServletRequest();
```
snoopdave, line 93: Direct static ActionContext lookups add thread-local and lifecycle coupling and make these actions harder to unit test. Struts 7 provides the replacement interfaces under org.apache.struts2.action—ServletRequestAware, ServletResponseAware, and ParametersAware—just as this PR correctly uses the new upload-aware API. Please use those injected APIs for this and the other nine new ActionContext.getContext() call sites.

Problem exists at these locations as well:

app/src/main/java/org/apache/roller/weblogger/planet/ui/PlanetGroupSubs.java:78
app/src/main/java/org/apache/roller/weblogger/planet/ui/PlanetGroups.java:55
app/src/main/java/org/apache/roller/weblogger/planet/ui/PlanetConfig.java:84
app/src/main/java/org/apache/roller/weblogger/ui/struts2/admin/GlobalConfig.java:102,162
app/src/main/java/org/apache/roller/weblogger/ui/struts2/admin/GlobalCommentManagement.java:247
app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/Members.java:61
app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/FolderEdit.java:83
app/src/main/java/org/apache/roller/weblogger/ui/struts2/core/Register.java:438

app/src/test/resources/roller-jettyrun.properties
```diff
@@ -49,3 +50,13 @@ cache.sitewide.enabled=false
 cache.weblogpage.enabled=false
 cache.weblogfeed.enabled=false
 cache.planet.enabled=false
+
+# Authentication defaults to Roller's own user database, which needs no extra
+# services. To develop against OIDC instead, start Keycloak with
```
snoopdave, line 55: not critical, but should OIDC stuff live in the stacked PR 155 that follows this?

PR #157, opened 2026-08-12
Replace the Selenium suite with Playwright browser tests run across all auth methods

mraible:
Stacked on #156 (3/3 in the stack); review only the top commit.

**Removed:** the it-selenium module, its Jetty/Derby test harness, and its CI steps. Its user-journey coverage is ported to the new it-playwright module below, so no test scenario is lost — the journey actually grows (media upload, OIDC, login-page checks). The GitHub Actions workflow is rewritten accordingly.

it-playwright holds Java Playwright tests that point at a running Roller and adapt to how it is configured:

- `NewUserJourneyIT` ports the old Selenium journey and extends it: register the first user, sign in, create a weblog, publish an entry, read it back on the blog, then enable file uploads from the server admin page (they ship disabled; the first user is the admin) and upload an image that is verified to be served back, and import an OPML file whose blogroll must appear
- `OidcLoginIT` signs in through the identity provider as an administrator and a regular user, verifies only the administrator reaches server administration, and creates a weblog and publishes an entry while signed in via OIDC
- `LoginPageIT` checks the login page offers exactly the sign-in mechanisms of the configured authentication method
- `WebServicesIT` is the only coverage of the servlets forked from javax-only libraries during the migration: it enables the APIs from the server admin page, lists weblogs and publishes an entry over Blogger/MetaWeblog XML-RPC, verifies the entry renders, and fetches the AtomPub service document over basic auth. It immediately caught a pre-existing bug where AtomPub basic auth always returned 401 (fixed in the Jakarta PR)

Each test skips whatever its instance does not offer, which keeps casual local runs friendly but could let a misconfigured instance pass with everything skipped. Passing `-Droller.expectedAuth=db|oidc|db-oidc` turns mismatches and unexpected skips into failures.

CI runs the suite three ways: db on Jetty with Derby, plus oidc and db-oidc against the Docker Compose stack through an `AUTHENTICATION_METHOD` matrix. The compose jobs drive Roller's auto-installer on the fresh database and upload Playwright traces on failure. All three legs were verified locally against fresh databases before the workflow was wired up.

it-playwright is deliberately not a reactor module since it needs a browser and a running instance; see its README for local usage. The ldap and cma methods stay uncovered because neither stack provides a directory server or container-managed realms. Failed tests record a trace to `target/playwright-traces`, viewable with `npx playwright show-trace`.


mbien:
what is the trigger for the test framework swap? A compatibility blocker or something else?

The text describes the what not the why atm.

mraible:
> what is the trigger for the test framework swap? A compatibility blocker or something else?

I've used Selenium in the past and found it flaky. I've had better luck with Cypress (esp in JHipster) and most recently with Playwright. Since I could pick any of them, I chose Playwright since it's where my most recent experience is.

PR #159, opened 2026-08-13
Use double-quoted attributes in media file links inserted into new entries

mraible:
The "create a post from your uploaded files" flow builds its image and enclosure markup with single-quoted attributes and an invalid closing </img> tag, and the resulting image does not render after publishing. The media chooser on the entry editor inserts the same link with double quotes and a self-closing img tag, and that form works, so this flow now matches it.

Reported by Greg Huber on dev@roller while testing the ROL-2183 branch; the bug exists on master independently of that work.

### DIFF
--- a/app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/EntryAddWithMediaFile.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/EntryAddWithMediaFile.java
@@ -76,14 +76,14 @@ public class EntryAddWithMediaFile extends MediaFileBase {
 
                     if (mediaFile.isImageFile()) {
                         link = "<p>" + mediaFile.getName() + "</p>";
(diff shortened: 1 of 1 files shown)


mbien (approved):

app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/EntryAddWithMediaFile.java
```diff
@@ -76,14 +76,14 @@ public String execute() {
 
                     if (mediaFile.isImageFile()) {
                         link = "<p>" + mediaFile.getName() + "</p>";
-                        link += "<a href='<url>'><img src='<urlt>' alt='<name>' width='<width>' height='<height>'></img></a>";
+                        link += "<a href=\"<url>\"><img src=\"<urlt>\" alt=\"<name>\" width=\"<width>\" height=\"<height>\" /></a>";
```
mbien, line 79: so the fix is the missing `/` of this line and the quote change is to match other generated code? I see usage of both `href='` and `href="` but double quote is more common so thats good.

PR #162, opened 2026-08-15
Scope authoring resource lookups to the action weblog

snoopdave:
## Summary

Roller's manager APIs expose two styles of lookup: by id, and by a weblog plus
some other key — `getTemplateByLink(Weblog, String)`,
`getTemplateByName(Weblog, String)` and friends. The authoring UI operates in
the context of a single weblog throughout, but has only the id-based form
available for several resource types, so the two styles are mixed within the
same action.

This adds weblog-scoped overloads beside the existing id-based lookups and
points the authoring actions at them, so the lookups an action performs are
consistently expressed in terms of the weblog it is working on. The new
methods follow the shape of the existing `getTemplateByLink(Weblog, String)`
rather than introducing a different convention.

## Manager methods added

| Manager | Method | Named query |
|---|---|---|
| `WeblogManager` | `getTemplate(Weblog, String)` | `WeblogTemplate.getByWeblog&Id` |
| `WeblogEntryManager` | `getWeblogEntry(Weblog, String)` | `WeblogEntry.getByWebsite&Id` |
| `WeblogEntryManager` | `getWeblogCategory(Weblog, String)` | `WeblogCategory.getByWeblog&Id` |
| `WeblogEntryManager` | `getComment(Weblog, String)` | `WeblogEntryComment.getByWebsite&Id` |
| `BookmarkManager` | `getBookmark(Weblog, String)` | `WeblogBookmark.getByWebsite&Id` |
| `BookmarkManager` | `getFolderById(Weblog, String)` | `WeblogBookmarkFolder.getByWebsite&Id` |
| `MediaFileManager` | `getMediaFile(Weblog, String)` | `MediaFile.getByWeblogAndId` |
| `MediaFileManager` | `getMediaFileDirectory(Weblog, String)` | `MediaFileDirectory.getByWeblogAndId` |

Each is backed by a named query filtering on the owning weblog, and returns
null when there is no match.

The folder lookup is named `getFolderById` rather than being an overload:
`getFolder(Weblog, String)` already exists as the by-name lookup, and the two
would otherwise share an erasure.

## Call sites

Updated across templates, entries, categories, comments, bookmarks, folders
and media files in `ui/struts2/editor/`. The action weblog is resolved by the
interceptor stack (`params` -> `UIActionInterceptor` -> `UISecurityInterceptor`
-> `UIActionPrepareInterceptor`) before `myPrepare()` runs, so
`getActionWeblog()` is available at each of these call sites without
reordering anything.

The two comment loops in `Comments` previously compared the weblog after
loading; they now use the scoped lookup instead, which also removes an NPE on
ids with no match.

## Not changed

`RollerResourceLoader` and `GlobalCommentManagement` have no single weblog in
context by design and keep the id-based lookups. `EntryBean`,
`StylesheetEdit`, `CommentDataServlet` and the XML-RPC and Atom handlers reach
their weblog by other means and were left alone.

## Tests

- `business/WeblogScopedLookupTest` — manager layer, covering each new overload
- `ui/struts2/editor/TemplateEditScopingTest` — action layer
- `ui/struts2/editor/AuthoringActionScopingTest` — action layer, covering
  `EntryRemove`, `EntryEdit`, `CategoryRemove` and `BookmarkEdit`

Full suite on JDK 11: **182 run, 0 failures, 0 errors, 1 skipped**.

### DIFF
--- a/app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/Bookmarks.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/Bookmarks.java
@@ -72,7 +72,7 @@ public class Bookmarks extends UIAction {
         try {
             BookmarkManager bmgr = WebloggerFactory.getWeblogger().getBookmarkManager();
             if (!StringUtils.isEmpty(getFolderId())) {
-                setFolder(bmgr.getFolder(getFolderId()));
+                setFolder(bmgr.getFolderById(getActionWeblog(), getFolderId()));
             } else {
                 setFolder(bmgr.getDefaultFolder(getActionWeblog()));
                 if (getFolder() != null) {
@@ -134,7 +134,7 @@ public class Bookmarks extends UIAction {
                     if (log.isDebugEnabled()) {
                         log.debug("Deleting bookmark - " + bookmarks[j]);
                     }
-                    bookmark = bmgr.getBookmark(bookmarks[j]);
+                    bookmark = bmgr.getBookmark(getActionWeblog(), bookmarks[j]);
                     if (bookmark != null) {
                         bmgr.removeBookmark(bookmark);
                     }
@@ -160,7 +160,7 @@ public class Bookmarks extends UIAction {
 
         try {
             BookmarkManager bmgr = WebloggerFactory.getWeblogger().getBookmarkManager();
-            WeblogBookmarkFolder fd = bmgr.getFolder(getFolderId());
+            WeblogBookmarkFolder fd = bmgr.getFolderById(getActionWeblog(), getFolderId());
 
             if (fd != null) {
 
@@ -203,7 +203,7 @@ public class Bookmarks extends UIAction {
         try {
             BookmarkManager bmgr = WebloggerFactory.getWeblogger().getBookmarkManager();
             if (!StringUtils.isEmpty(viewFolderId)) {
-                setFolder(bmgr.getFolder(viewFolderId));
+                setFolder(bmgr.getFolderById(getActionWeblog(), viewFolderId));
                 setFolderId(viewFolderId);
             }
         } catch (WebloggerException ex) {
@@ -225,12 +225,12 @@ public class Bookmarks extends UIAction {
             }
 
             // Move bookmarks to new parent folder.
-            WeblogBookmarkFolder newFolder = bmgr.getFolder(getTargetFolderId());
+            WeblogBookmarkFolder newFolder = bmgr.getFolderById(getActionWeblog(), getTargetFolderId());
             String bookmarks[] = getSelectedBookmarks();
 
             if (null != bookmarks && bookmarks.length > 0) {
                 for (int j = 0; j < bookmarks.length; j++) {
-                    WeblogBookmark bd = bmgr.getBookmark(bookmarks[j]);
+                    WeblogBookmark bd = bmgr.getBookmark(getActionWeblog(), bookmarks[j]);
                     newFolder.addBookmark(bd);
                     bd.setFolder(newFolder);
                     bmgr.saveBookmark(bd);
--- a/app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/Bookmarks.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/Bookmarks.java
@@ -228,9 +228,22 @@ public class Bookmarks extends UIAction {
             WeblogBookmarkFolder newFolder = bmgr.getFolderById(getActionWeblog(), getTargetFolderId());
             String bookmarks[] = getSelectedBookmarks();
 
+            if (newFolder == null) {
+                addError("bookmarksForm.error.move");
+                return execute();
+            }
+
             if (null != bookmarks && bookmarks.length > 0) {
-                for (int j = 0; j < bookmarks.length; j++) {
-                    WeblogBookmark bd = bmgr.getBookmark(getActionWeblog(), bookmarks[j]);
+                List<WeblogBookmark> bookmarksToMove = new ArrayList<>();
+                for (String bookmarkId : bookmarks) {
+                    WeblogBookmark bd = bmgr.getBookmark(getActionWeblog(), bookmarkId);
+                    if (bd == null) {
+                        addError("bookmarksForm.error.move");
+                        return execute();
+                    }
+                    bookmarksToMove.add(bd);
+                }
+                for (WeblogBookmark bd : bookmarksToMove) {
                     newFolder.addBookmark(bd);
                     bd.setFolder(newFolder);
                     bmgr.saveBookmark(bd);
--- a/app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/CategoryRemove.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/CategoryRemove.java
@@ -113,6 +113,10 @@ public class CategoryRemove extends UIAction {
 
                 if (getTargetCategoryId() != null) {
                     WeblogCategory target = wmgr.getWeblogCategory(getActionWeblog(), getTargetCategoryId());
+                    if (target == null) {
+                        addError("categoryForm.notFound");
+                        return execute();
+                    }
                     wmgr.moveWeblogCategoryContents(getCategory(), target);
                     WebloggerFactory.getWeblogger().flush();
                 }
--- a/app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/CategoryRemove.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/CategoryRemove.java
@@ -73,7 +73,7 @@ public class CategoryRemove extends UIAction {
         try {
             WeblogEntryManager wmgr = WebloggerFactory.getWeblogger().getWeblogEntryManager();
             if(!StringUtils.isEmpty(getRemoveId())) {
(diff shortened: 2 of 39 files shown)


mbien (approved):
makes sense

mraible (commented):
Reviewed. The scoped manager changes follow existing project patterns. Please address the null-handling regressions noted in the review before merge.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/WeblogConfig.java
```diff
@@ -136,7 +136,7 @@ public String save() {
                 // if blogger category changed then lookup new cat and set it
                 if(getBean().getBloggerCategoryId() != null &&
                         !weblog.getBloggerCategory().getId().equals(getBean().getBloggerCategoryId())) {
-                    weblog.setBloggerCategory(wmgr.getWeblogCategory(getBean().getBloggerCategoryId()));
+                    weblog.setBloggerCategory(wmgr.getWeblogCategory(getActionWeblog(), getBean().getBloggerCategoryId()));
```
mraible: This one persists bad data: when `bloggerCategoryId` doesn't resolve inside the action weblog, the scoped lookup returns `null` and `setBloggerCategory(null)` is committed (bloggercatid is nullable). From then on line 138 dereferences `weblog.getBloggerCategory()` on every save, inside a catch that only handles `WebloggerException`, so the Settings page for that weblog 500s until someone repairs the row. A `null` check that adds a validation error and leaves the existing category alone closes it.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/MediaFileBase.java
```diff
@@ -62,7 +62,7 @@ protected void doDeleteMediaFile() {
             log.debug("Processing delete of file id - " + this.mediaFileId);
             MediaFileManager manager = WebloggerFactory.getWeblogger()
                     .getMediaFileManager();
-            MediaFile mediaFile = manager.getMediaFile(this.mediaFileId);
+            MediaFile mediaFile = manager.getMediaFile(getActionWeblog(), this.mediaFileId);
```
mraible: `doDeleteMediaFile` here, `doIncludeMediaFileInGallery` (line 87), and `doMoveSelected` (line 150, `targetDirectory`) all pass the now-nullable scoped result straight through: `removeMediaFile(weblog, null)` ends in an `IllegalArgumentException` from `em.remove(null)`, the others NPE, and the catch only covers `WebloggerException`. A foreign or stale id now yields a 500 page instead of the `mediaFile`.`delete.error` / `includeInGallery`.error / `move.errors` messages the code intends.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/Bookmarks.java
```diff
@@ -225,12 +225,12 @@ public String move() {
...
             // Move bookmarks to new parent folder.
-            WeblogBookmarkFolder newFolder = bmgr.getFolder(getTargetFolderId());
+            WeblogBookmarkFolder newFolder = bmgr.getFolderById(getActionWeblog(), getTargetFolderId());
             String bookmarks[] = getSelectedBookmarks();
 
             if (null != bookmarks && bookmarks.length > 0) {
                 for (int j = 0; j < bookmarks.length; j++) {
-                    WeblogBookmark bd = bmgr.getBookmark(bookmarks[j]);
+                    WeblogBookmark bd = bmgr.getBookmark(getActionWeblog(), bookmarks[j]);
                     newFolder.addBookmark(bd);
```
mraible: `move()` never `null`-checks `newFolder` (line 228) or bd (line 233), although `delete()` at line 138 does. A `targetFolderId` from another weblog gives `newFolder == null`, `addBookmark` NPEs past the `WebloggerException` catch, and the user gets a 500 rather than `bookmarksForm`.`error.move`. An unresolved `selectedBookmarks` entry also adds a `null` to the folder's in-memory list before bd.`setFolder` blows up.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/MediaFileEdit.java
```diff
@@ -67,7 +67,7 @@ public void myPrepare() {
         try {
             MediaFileManager mgr = WebloggerFactory.getWeblogger().getMediaFileManager();
             if (!StringUtils.isEmpty(bean.getDirectoryId())) {
-                setDirectory(mgr.getMediaFileDirectory(bean.getDirectoryId()));
+                setDirectory(mgr.getMediaFileDirectory(getActionWeblog(), bean.getDirectoryId()));
             }
         } catch (WebloggerException ex) {
             log.error("Error looking up media file directory", ex);
```
mraible: The directory set here can now be `null` for a `directoryId` outside the action weblog, and `myValidate` (line 82) calls `getDirectory()`.`getMediaFile(...)` on it. `save()` runs `myValidate` before entering its try, so a foreign or deleted directory id is an unhandled NPE instead of INPUT with a validation error.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/EntryEdit.java
```diff
@@ -104,7 +104,7 @@ public void myPrepare() {
                 // retrieve from DB WeblogEntry based on ID
                 WeblogEntryManager wmgr = WebloggerFactory.getWeblogger()
                         .getWeblogEntryManager();
-                setEntry(wmgr.getWeblogEntry(getBean().getId()));
+                setEntry(wmgr.getWeblogEntry(getActionWeblog(), getBean().getId()));
             } catch (WebloggerException ex) {
                 log.error(
                         "Error looking up entry by id - " + getBean().getId(),
```
mraible: `myPrepare` now leaves entry `null` for any id outside the action weblog (the new test asserts exactly that), but `execute()` (line 126) hands it to `EntryBean`.`copyFrom`, which calls `entry.getId()` immediately. A co-author deleting an entry you have open, or a crafted `bean.id`, is now a stack-trace page rather than a not-found message. `TemplateEdit` already handles the same `null` with 'Unable to locate specified template'; `CategoryEdit`.execute, `FolderEdit`.`myValidate`, and `BookmarkEdit`.`myValidate` (line 123) have the same gap.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/Comments.java
```diff
@@ -102,7 +102,7 @@ public void loadComments() {
 
             // lookup weblog entry if necessary
             if (!StringUtils.isEmpty(getBean().getEntryId())) {
-                setQueryEntry(wmgr.getWeblogEntry(getBean().getEntryId()));
+                setQueryEntry(wmgr.getWeblogEntry(getActionWeblog(), getBean().getEntryId()));
```
mraible: Different failure shape here: a foreign or stale `entryId` now makes `queryEntry` `null`, and `null` silently means 'all comments in the weblog', so the listing widens from one entry to the whole blog without any indication. Probably worth treating an unresolved `entryId` as an empty result or an error rather than as no filter.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/MediaFileImageDim.java
```diff
@@ -50,7 +50,7 @@ public MediaFileImageDim() {
     public String execute() {
         try {
             MediaFileManager mgr = WebloggerFactory.getWeblogger().getMediaFileManager();
-            MediaFile mediaFile = mgr.getMediaFile(getMediaFileId());
+            MediaFile mediaFile = mgr.getMediaFile(getActionWeblog(), getMediaFileId());
             bean.copyFrom(mediaFile);
```
mraible: Same family: the scoped lookup can return `null` and `bean.copyFrom(mediaFile)` dereferences it inside a catch that only handles `WebloggerException`.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/EntryAddWithMediaFile.java
```diff
@@ -71,7 +71,7 @@ public String execute() {
 
             if (selectedImages != null) {
                 for (String image : selectedImages) {
-                    MediaFile mediaFile = manager.getMediaFile(image);
+                    MediaFile mediaFile = manager.getMediaFile(getActionWeblog(), image);
                     String link;
 
                     if (mediaFile.isImageFile()) {
```
mraible: A selected image outside the action weblog now resolves to `null` and the loop dereferences it; the generic catch swallows the NPE, so the new-entry form silently loses the media links rather than telling the user what happened.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/CategoryRemove.java
```diff
@@ -112,7 +112,7 @@ public String remove() {
                 WeblogEntryManager wmgr = WebloggerFactory.getWeblogger().getWeblogEntryManager();
 
                 if (getTargetCategoryId() != null) {
-                    WeblogCategory target = wmgr.getWeblogCategory(getTargetCategoryId());
+                    WeblogCategory target = wmgr.getWeblogCategory(getActionWeblog(), getTargetCategoryId());
                     wmgr.moveWeblogCategoryContents(getCategory(), target);
```
mraible: `targetCategoryId` outside the action weblog gives a `null` target here, which goes straight into `moveWeblogCategoryContents`.

PR #163, opened 2026-08-15
Remove legacy outbound trackback action

snoopdave:
## Summary

- Remove the legacy outbound trackback action and its supporting code.
- Remove the obsolete action configuration, editor remnants, and the
  now-unreferenced configuration property that went with it.
- Add regression coverage confirming that the entry actions expose only their
  current methods.

## Rationale

The editor stopped presenting outbound trackback controls some time ago, but
the action and its supporting code remained in the application with nothing
reaching them. This removes that dormant path. Incoming trackback handling is
unchanged.

## User impact

Manual outbound trackback sending from the entry editor is no longer
available. Entry creation, entry editing, and incoming trackbacks otherwise
behave as before.

## Validation

- `mvn -pl app clean verify -Dnet.bytebuddy.experimental=true` — 160 tests,
  0 failures, 1 skipped.
- Confirmed the packaged JAR no longer contains the removed classes.


app/src/test/java/org/apache/roller/weblogger/ui/struts2/editor/EntryTrackbackRemovalTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NodeList;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class EntryTrackbackRemovalTest {
```
mbien, line 34: nitpick: regression tests for code removal are probably not super useful, no? (i know LLMs love to generate tests for _everything_)

mbien (approved):
makes sense

app/src/test/java/org/apache/roller/weblogger/ui/struts2/editor/EntryTrackbackRemovalTest.java
```diff
@@ -0,0 +1,76 @@
...
+            Set<String> expectedActions = Set.of("entryAdd", "entryEdit");
+            int actionsChecked = 0;
+
+            for (int i = 0; i < actions.getLength(); i++) {
+                Element action = (Element) actions.item(i);
+                if (expectedActions.contains(action.getAttribute("name"))) {
+                    NodeList allowedMethods = action.getElementsByTagName("allowed-methods");
+                    assertEquals(1, allowedMethods.getLength());
+                    String methods = allowedMethods.item(0).getTextContent();
+                    assertFalse(Set.of(methods.trim().split("\\s*,\\s*")).contains("trackback"));
```
mraible: Tiny one: `Set.of` throws `IllegalArgumentException` on duplicate elements, so a duplicated method name in allowed-methods would fail this test with "duplicate element" instead of an assertion about trackback. `List.of(...)`.`contains(...)` does the same job without that trap.

mraible (approved):
Reviewed with the same multi-agent find-and-verify pass we've been using on the ROL-2183 stack. No runtime defects: the action, the allowed-method, and the JSP remnant are all gone and nothing reaches the removed path. Two leftovers outside the diff worth sweeping in a follow-up or here:

- `docs/roller-user-guide.adoc` still has the "Sending trackbacks" section (around lines 788-811) telling users to scroll down and click Send Trackback.
- The trackback i18n keys (`error.trackback`*, `weblogEdit`.trackback*, `weblogEdit`.`sendTrackback`, `weblogEdit`.`trackbackUrl` and the success/failure/error variants) survive in `ApplicationResources.properties` and the de, es, fr, ja, ko, ru, and zh_CN bundles, plus the div.`trackbackUrl` rule in `themes/base.css`. Nothing renders them anymore.

One small inline note on the regression test. Approving.

PR #164, opened 2026-08-29
Check weblog permission in XML-RPC Blogger/MetaWeblog handlers

snoopdave:
The Blogger and MetaWeblog XML-RPC handlers are Roller's legacy remote-publishing
API. This change makes them apply the same per-weblog and per-entry permission
model that the rest of Roller uses, so a caller is authorized against the weblog
or entry an operation actually touches.

## What changed

- Apply an explicit per-method permission matrix (weblog `ADMIN` / `EDIT_DRAFT` /
  `POST`, or entry-level write permission) to the weblog or entry actually
  touched.
- `getUserInfo` returns only the authenticated user; `getUsersBlogs` returns only
  API-enabled member weblogs.
- Entry-ID operations authorize against the entry's own weblog.
- Foreign and unknown identifiers produce the same fault.
- Bound the number of entries a recent-posts call scans for a limited member.

## Tests

Table-driven integration suite over every exposed handler method: authorized
member, authenticated non-member, insufficient role, disabled user/weblog/API,
draft-to-published transition, filtered recent posts, category/weblog mismatch,
and foreign/absent resource identifiers.


app/src/main/java/org/apache/roller/weblogger/webservices/xmlrpc/BaseAPIHandler.java
```diff
@@ -102,97 +105,112 @@ public BaseAPIHandler() {
...
+                && !entry.getWebsite().hasUserPermission(user, additionalAction)) {
+            return null;
+        }
+        return entry;
+    }
+
+    private boolean isWeblogAvailable(Weblog website) {
+        return website != null
+                && Boolean.TRUE.equals(website.getVisible())
+                && Boolean.TRUE.equals(website.getActive());
```
mraible: `Weblog.active` isn't a disabled flag: per its javadoc (and ROL-485) it's the user-settable "include in front page / planet listings" toggle, and the web UI never blocks editing an inactive weblog. Gating here means an owner who unticks Active in Weblog Settings to hide the blog from the front page loses every XML-RPC call for it (`newPost`, `editPost`, `getRecentPosts`, and via `getEntryForWrite` even `deletePost` by id) with "not found or disabled". The pre-PR check was visible only, which is the flag that means what this code wants.

app/src/main/java/org/apache/roller/weblogger/webservices/xmlrpc/BloggerAPIHandler.java
```diff
@@ -463,21 +468,35 @@ public Object getRecentPosts(String appkey, String blogid, String userid,
...
             Weblogger roller = WebloggerFactory.getWeblogger();
             WeblogEntryManager weblogMgr = roller.getWeblogEntryManager();
             if (weblog != null) {
                 WeblogEntrySearchCriteria wesc = new WeblogEntrySearchCriteria();
                 wesc.setWeblog(weblog);
                 wesc.setEndDate(new Date());
+                if (weblog.hasUserPermission(user, WeblogPermission.POST)) {
+                    wesc.setMaxResults(numposts);
+                } else {
+                    wesc.setMaxResults(Math.max(numposts, DRAFT_SCAN_CAP));
```
mraible: `Math.max(numposts, DRAFT_SCAN_CAP)` makes the cap a floor, not a ceiling: `numposts=1000000` scans a million rows, and for a normal numposts a limited member only ever sees drafts that fall within the 200 most recent entries. With more than 200 published entries newer than their draft, `getRecentPosts` returns an empty list even though `getPost` on the draft's id succeeds. Same shape at `MetaWeblogAPIHandler.java`:445. `Math.min` is what the javadoc describes, and the drafts case probably needs a status-aware query rather than a scan of everything.

app/src/main/java/org/apache/roller/weblogger/webservices/xmlrpc/BaseAPIHandler.java
```diff
@@ -102,97 +105,112 @@ public BaseAPIHandler() {
...
+        return validateWeblog(blogid, user, requiredAction);
+    }
+
+    /**
+     * Validate a weblog for an already authenticated user.
+     */
+    protected Weblog validateWeblog(String blogid, User user,
+            String requiredAction) throws Exception {
+        WeblogManager weblogMgr = WebloggerFactory.getWeblogger().getWeblogManager();
+        Weblog website = weblogMgr.getWeblogByHandle(blogid);
```
mraible: The deleted `validate()` wrapped the user and weblog lookups in a try/catch that logged and converted backend failures into an authorization fault. Here `getWeblogByHandle` runs outside any try, so a `WebloggerException` (transient DB trouble, say) escapes as a raw exception and the XML-RPC servlet returns a generic server fault carrying the internal message. Easy to trigger in tests with a hyphenated blogid. Worth restoring the wrap.

app/src/main/java/org/apache/roller/weblogger/webservices/xmlrpc/MetaWeblogAPIHandler.java
```diff
@@ -330,14 +333,8 @@ public Object getPost(String postid, String userid, String password)
...
-        Weblogger roller = WebloggerFactory.getWeblogger();
-        WeblogEntryManager weblogMgr = roller.getWeblogEntryManager();
-        WeblogEntry entry = weblogMgr.getWeblogEntry(postid);
-        
-        if (entry == null) {
-            throw new XmlRpcException(INVALID_POSTID, INVALID_POSTID_MSG);
-        }
-        validate(entry.getWebsite().getHandle(), userid, password);
+        User user = validateUser(userid, password);
+        WeblogEntry entry = validateEntry(postid, user, null);
```
mraible, line 339: Intended? `getPost` used to require only weblog membership; `validateEntry(postid, user, null)` now requires `entry.hasWritePermissions(user)`, which is false for `EDIT_DRAFT` members on published entries. So a limited member can no longer read, via the API, a published entry that the site serves publicly, and their client reports the post as missing. The new test at `XMLRPCWeblogPermissionTest.java`:282 locks this in, so if it's not deliberate it's worth catching before it becomes the contract.

app/src/main/java/org/apache/roller/weblogger/webservices/xmlrpc/BaseAPIHandler.java
```diff
@@ -102,97 +105,112 @@ public BaseAPIHandler() {
...
+    /**
+     * Returns an entry only when it belongs to an available XML-RPC weblog and
+     * the user may edit it. An optional additional weblog action can be
+     * required for transitions such as publishing.
+     */
+    protected WeblogEntry validateEntry(String postid, User user,
+            String additionalAction) throws Exception {
+        WeblogEntry entry = getEntryForWrite(postid, user, additionalAction);
+        if (entry == null) {
+            throw new XmlRpcException(INVALID_POSTID, INVALID_POSTID_MSG);
```
mraible: I read the description's "foreign and unknown identifiers produce the same fault" as deliberate anti-enumeration, which makes sense for entries the caller can't see. This also covers the case where the caller can see and edit the entry and only lacks POST: a limited member who just saved a draft with publish=false retries with publish=true and gets INVALID_POSTID for the id that worked seconds ago. Clients that treat that fault as "deleted on the server" will drop or re-create the post. For an entry the caller already has access to, a not-authorized fault leaks nothing and is far kinder.

app/src/main/java/org/apache/roller/weblogger/webservices/xmlrpc/BloggerAPIHandler.java
```diff
@@ -463,21 +468,35 @@ public Object getRecentPosts(String appkey, String blogid, String userid,
...
         mLogger.debug("     Number: " + numposts);
         
-        Weblog weblog = validate(blogid, userid,password);
+        User user = validateUser(userid, password);
+        Weblog weblog = validateWeblog(blogid, user,
+                WeblogPermission.EDIT_DRAFT);
         
         try {
             Vector<Object> results = new Vector<>();
+            if (numposts <= 0) {
```
mraible: Intended? The old Blogger `getRecentPosts` ignored numposts and returned everything; this returns an empty list for numposts <= 0 and truncates otherwise. Honoring numposts is arguably the correct behavior, but it's a silent contract change for Blogger 1.0 clients and export scripts that relied on the old one, so it deserves a line in the description at least. Returning everything for numposts <= 0 (rather than nothing) would keep the old tooling working.

app/src/main/java/org/apache/roller/weblogger/webservices/xmlrpc/BaseAPIHandler.java
```diff
@@ -102,97 +105,112 @@ public BaseAPIHandler() {
...
+        // Use one response for missing, unavailable, and inaccessible weblogs.
+        if (!isWeblogAvailable(website)
+                || !website.hasUserPermission(user, requiredAction)) {
             throw new XmlRpcNotAuthorizedException(WEBLOG_DISABLED_MSG);
         }
-        if ( !weblogFound ) {
-            throw new XmlRpcException(WEBLOG_NOT_FOUND, WEBLOG_NOT_FOUND_MSG);
-        }
-        if ( !apiEnabled ) {
+        if (!Boolean.TRUE.equals(website.getEnableBloggerApi())) {
```
mraible: Small one: the `enableBloggerApi` check lives here and again in `getEntryForWrite` (line 197), and the two paths report it differently (`BLOGGERAPI_DISABLED` here, collapsed into `INVALID_POSTID` there). Doing it once in `isWeblogAvailable` would keep the faults consistent.

app/src/main/java/org/apache/roller/weblogger/webservices/xmlrpc/MetaWeblogAPIHandler.java
```diff
@@ -82,7 +83,8 @@ public Object getCategories(String blogid, String userid, String password)
         mLogger.debug("     BlogId: " + blogid);
```
mraible, line 85: Consistency note rather than a defect: XML-RPC now grants `EDIT_DRAFT` members read access (`getCategories`, `getRecentPosts`, `getPost`) while `RollerAtomHandler`.`canView`/`canEdit` requires POST on the same weblog. Whichever is right, the two remote APIs probably want the same answer, and the Atom one is the older reference.

mraible (commented):
Reviewed. The per-method behavior is moving in the right direction. Please resolve the regression notes and document the intentional compatibility changes before merge.

app/src/main/java/org/apache/roller/weblogger/webservices/xmlrpc/BaseAPIHandler.java
```diff
@@ -102,97 +105,133 @@ public BaseAPIHandler() {
...
+
+        if (!Boolean.TRUE.equals(user.getEnabled())) {
             throw new XmlRpcNotAuthorizedException(USER_DISABLED_MSG);
         }
-        if ( !weblogEnabled ) {
-            throw new XmlRpcNotAuthorizedException(WEBLOG_DISABLED_MSG);
+
+        if (!WebloggerRuntimeConfig.getBooleanProperty("webservices.enableXmlRpc")) {
+            throw new XmlRpcNotAuthorizedException(BLOGGERAPI_DISABLED_MSG);
         }
```
mbien: fail-fast: should this check be moved before the try-block? The user enabled check could be also moved before the authentication query (`getPasswordEncoder().matche()`).

mbien (approved):

PR #165, opened 2026-08-29
Use the current Roller session during OAuth authorization

snoopdave:
The OAuth 1.0a consent step should authorize the user who is signed in, the way the rest of Roller resolves identity from the session. This change moves it onto that model and keeps request-token approval atomic.

## What changed

- Derive the authorizing identity from the Roller session and require an enabled user.
- Save pending consent as a normal Spring login continuation when no session is present.
- Keep the bound-consumer check and tolerate legacy identity parameters only when they agree with the session.
- Approve the exact pending token with one conditional update; a same-user retry is idempotent, while another user or an exchanged token is refused.
- Re-render an expired or already-used consent form with a fresh salt.
- Handle missing or expired token state consistently with `oauth_problem=permission_denied` instead of a container error.
- Keep the existing `markAsAuthorized` manager method for compatibility.

PR #167 remains the merge prerequisite. This branch reuses its `SaltValidator` implementation byte-for-byte, so that shared file drops from this PR after #167 lands.

## Tests

- `AuthorizationServletTest`: 13 passing cases for session identity, refusal behavior, login continuation, missing state, and stale consent forms.
- `JPAOAuthManagerTest`: 2 passing persistence cases, including same-user retry and different-user refusal.
- Clean JDK 11 reactor: 172 tests, 0 failures, 0 errors, 1 skipped.
- Disposable #165 + #167 integration merge: 35 focused tests passed.


app/src/main/java/org/apache/roller/weblogger/webservices/oauth/AuthorizationServlet.java
```diff
@@ -71,40 +81,93 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+                    accessor.consumer.consumerKey, accessor.requestToken, userId)) {
+                denyPermission(response);
+                return;
+            }
+            WebloggerFactory.getWeblogger().flush();
+
+            accessor.setProperty("userId", userId);
+            accessor.setProperty("authorized", Boolean.TRUE);
+
             returnToConsumer(request, response, accessor);
```
mraible: This is where the fixation attack survives. Attacker starts the consumer flow, gets request token T, and sends a logged-in victim `/roller-services/oauth/authorize?oauth_token=T`. The victim clicks Authorize, this conditional update binds T to the victim, and the attacker finishes the exchange at the consumer: `AccessTokenServlet` checks only `authorized == TRUE`, so it hands out an access token acting as the victim. The session check and one-shot approval don't help because the attacker never posts here. OAuth 1.0a fixed exactly this with `oauth_verifier`: generate one on approval, store it on the accessor, return it to the consumer through the callback, and require it on the access-token exchange.

app/src/main/java/org/apache/roller/weblogger/webservices/oauth/AuthorizationServlet.java
```diff
@@ -71,40 +81,93 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+        response.setStatus(HttpServletResponse.SC_FORBIDDEN);
+        response.setContentType("text/plain");
+        try (PrintWriter out = response.getWriter()) {
+            out.println("oauth_problem=" + PERMISSION_DENIED);
+        }
+    }
+
     private void sendToAuthorizePage(HttpServletRequest request, 
             HttpServletResponse response, OAuthAccessor accessor)
     throws IOException, ServletException{
```
mraible: `returnToConsumer` (line 190) redirects to the `oauth_callback` request parameter, which is attacker-controllable and seeded into the consent form's hidden field from the original GET link. A victim sent ...authorize?`oauth_token`=T&`oauth_callback`=`https://evil.example/` gets 302'd there with the freshly authorized token on the query string, and for an already-authorized T the GET path redirects with no click at all. Pre-existing, but this PR reworks the success path, so it's the moment to validate the callback against the consumer's registered one.

app/src/main/java/org/apache/roller/weblogger/webservices/oauth/AuthorizationServlet.java
```diff
@@ -71,40 +81,93 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
-            if (userId == null) {
-                // no user associted with the key, must be site-wide key,
-                // so get user to login and do the authorization process
+
+            // The approving identity comes from the browser session, consistent
+            // with the rest of the UI. Without a session there is nobody to
+            // approve on behalf of, so send the caller through the login flow.
+            User user = getAuthenticatedUser(request);
+            if (user == null) {
                 sendToAuthorizePage(request, response, accessor);
```
mraible: The comment says a session-less request goes through the login flow, but `sendToAuthorizePage` forwards to `/roller-ui/oauthAuthorize.rol`, which isn't a Spring-protected URL: UISecurityInterceptor returns DENIED and the user lands on the access-denied tile with no way to log in or resume. A user whose login expired mid-consent, or a first-time consent with no session, dead-ends there and the consumer's request token is stranded. A redirect to the login page with a saved request would do what the comment describes.

app/src/main/java/org/apache/roller/weblogger/business/jpa/JPAOAuthManagerImpl.java
```diff
@@ -139,6 +140,25 @@ public void markAsAuthorized(OAuthAccessor accessor, String userId)
...
+        if (consumerKey == null || requestToken == null || userName == null) {
+            return false;
+        }
+        try {
+            Query q = strategy.getNamedUpdate("OAuthAccessorRecord.authorizeRequestToken");
+            q.setParameter(1, userName);
+            q.setParameter(2, new Timestamp(new Date().getTime()));
+            q.setParameter(3, consumerKey);
+            q.setParameter(4, requestToken);
+            return q.executeUpdate() == 1;
```
mraible, line 155: `markAsAuthorized` was idempotent; this update requires authorized to be `null` or false, so approving a token that's already authorized but not yet exchanged affects zero rows and the servlet answers a bare 403 `permission_denied` instead of returning the user to the consumer. That happens on a retried or double-submitted approval. Treating already-authorized-by-the-same-user as success keeps the one-shot guarantee without breaking retries.

app/src/main/java/org/apache/roller/weblogger/webservices/oauth/AuthorizationServlet.java
```diff
@@ -42,7 +45,14 @@
...
+    /**
+     * One response for every refusal, so the endpoint reveals nothing about
+     * tokens the caller does not hold.
+     */
+    private static final String PERMISSION_DENIED = "permission_denied";
+
+
     @Override
     public void doGet(HttpServletRequest request, HttpServletResponse response)
             throws IOException, ServletException {
```
mraible, line 60: `doGet` still dereferences accessor without the `null` guard `doPost` gained (line 66): GET ...authorize?`oauth_consumer_key`=<key> before any request token exists NPEs on accessor.`getProperty` and surfaces as a container 500, where `doPost` now returns the uniform `permission_denied`.

app/src/main/webapp/WEB-INF/web.xml
```diff
@@ -153,6 +153,16 @@
...
 	</filter-mapping>
 
+	<!-- The OAuth consent form is served by a forward into /roller-ui/, so it
+	     already receives a token from LoadSaltFilter above, but it posts back
+	     outside /roller-ui/. Validate that one exact URL; the request-token and
+	     access-token endpoints are consumer-to-server calls carrying an OAuth
+	     signature and must not be included. -->
+	<filter-mapping>
+		<filter-name>ValidateSaltFilter</filter-name>
+		<url-pattern>/roller-services/oauth/authorize</url-pattern>
```
mraible: Good to gate the consent POST, but `ValidateSaltFilter` throws `ServletException("Security Violation")` when the salt is missing, expired (`SaltCache` entries live 60 minutes, and `cache.salt.size`=5000 evicts), already consumed, or issued to a different user. A user who leaves the consent page open for an hour and then clicks Authorize gets the container's 500 page rather than the consent form again or an OAuth problem response. Same on a cluster without a shared `SaltCache`. Catching that case in the servlet (re-render with a fresh salt) would keep the protection without the cliff.

app/src/main/java/org/apache/roller/weblogger/webservices/oauth/AuthorizationServlet.java
```diff
@@ -71,40 +81,93 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+            }
+            WebloggerFactory.getWeblogger().flush();
+
+            accessor.setProperty("userId", userId);
+            accessor.setProperty("authorized", Boolean.TRUE);
+
             returnToConsumer(request, response, accessor);
-            
+
+        } catch (OAuthProblemException e) {
```
mraible, line 168: Previously OAuthProblemException from `getAccessor` (`token_expired`, `token_rejected`) went through OAuthServlet.`handleException`, which sends the problem-specific status and a `WWW-Authenticate`: OAuth realm header. This catch collapses everything to a bare 403 with no realm header, while `doGet` on the same endpoint still reports the old way, so the same token state is described two different ways depending on method.

mraible (commented):
Reviewed. The session-binding, one-time approval, and validation changes are moving in the right direction. Please address the compatibility and error-handling notes before merge. Coordination with #154 remains necessary because both changes touch the same flow.

app/src/main/java/org/apache/roller/weblogger/business/OAuthManager.java
```diff
@@ -88,10 +88,39 @@ OAuthAccessor getAccessor(OAuthMessage requestMessage)
...
+     *
+     * @param consumerKey  key of the consumer the token was issued to
+     * @param requestToken the pending request token being approved
+     * @param userName     the approving user
+     * @return true if this call performed the approval; false if the record
+     *         did not match, was already authorized, or was already exchanged.
+     *         Callers should not distinguish these cases to the client.
+     * @throws OAuthException on persistence failure
+     */
+    boolean authorizeRequestToken(String consumerKey, String requestToken, String userName)
```
mraible, line 123: Minor: adding an abstract method to this interface while keeping `markAsAuthorized` deprecated 'for callers outside the project' is a bit contradictory; if external implementations are a concern, a default method covers them, and if they aren't, `markAsAuthorized` can just go.

app/src/test/java/org/apache/roller/weblogger/webservices/oauth/AuthorizationServletTest.java
```diff
@@ -0,0 +1,325 @@
...
+
+            factory.when(WebloggerFactory::getWeblogger).thenReturn(weblogger);
+            session.when(() -> RollerSession.getRollerSession(request)).thenReturn(rollerSession);
+            when(rollerSession.getAuthenticatedUser()).thenReturn(user("alice"));
+
+            when(request.getParameter("userId")).thenReturn("admin");
+
+            servlet.doPost(request, response);
+
+            verifyNothingAuthorizedFor("admin");
```
mraible: The mismatched-identity tests assert that nothing was authorized for admin, but not the 403, and not that alice wasn't authorized either. As written they'd pass if the mismatch were silently accepted for alice. Asserting the response status and the absence of any authorization pins the behavior the description promises.

app/src/main/java/org/apache/roller/weblogger/ui/core/filters/SaltValidator.java
```diff
@@ -0,0 +1,108 @@
...
+     * @return true when no Roller session is present or the submitted salt is valid
+     */
+    public static boolean consumeSubmittedSalt(HttpServletRequest request) {
+        RollerSession rollerSession = RollerSession.getRollerSession(request);
+        if (rollerSession == null) {
+            return true;
+        }
+
+        String userId = rollerSession.getAuthenticatedUser() != null
+                ? rollerSession.getAuthenticatedUser().getId() : "";
```
mbien: userId query can be moved after `salt==null` check + extract `rollerSession.getAuthenticatedUser()` into variable to call it only once.

mbien (approved):

PR #167, opened 2026-08-29
Separate submitted and response salts in UI filters

snoopdave:
Roller's UI request handling uses two salt filters: one processes the token submitted with a request, and the other generates the token for the next response. This change gives each filter a single responsibility and corrects their execution order.

## What changed

- Validate and consume the `salt` value submitted with the request.
- Run request validation before response-salt generation and create a fresh response salt only after successful processing.
- Route multipart `*.rol` validation through an interceptor that runs immediately after the upload interceptor.
- Remove the `salt.ignored.urls` setting and its `isIgnoredURL()` helper, whose shipped values did not match its behavior.

## Tests

- Cover missing, valid, and repeated submitted salts.
- Cover filter ordering and response-token generation.
- Cover multipart token handling on the supported Tomcat deployment and repository Jetty test setup.

### DIFF
--- a/app/src/main/java/org/apache/roller/weblogger/ui/core/filters/SaltValidator.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/core/filters/SaltValidator.java
@@ -1,0 +1,91 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.  For additional information regarding
+ * copyright in this work, please see the NOTICE file in the top level
+ * directory of this distribution.
+ */
+
+package org.apache.roller.weblogger.ui.core.filters;
+
+import java.util.Locale;
+import java.util.Objects;
+
+import javax.servlet.http.HttpServletRequest;
+
+import org.apache.roller.weblogger.ui.core.RollerSession;
+import org.apache.roller.weblogger.ui.rendering.util.cache.SaltCache;
+
+/**
+ * Shared validation for salts submitted by UI forms.
+ */
+public final class SaltValidator {
+
+    private static final String MULTIPART_FORM_DATA = "multipart/form-data";
+
+    private SaltValidator() {
+    }
+
+    /**
+     * Validates and consumes the salt submitted as a request parameter.
+     *
+     * @param request current request
+     * @return true when no Roller session is present or the submitted salt is valid
+     */
+    public static boolean consumeSubmittedSalt(HttpServletRequest request) {
+        RollerSession rollerSession = RollerSession.getRollerSession(request);
+        if (rollerSession == null) {
+            return true;
+        }
+
+        String userId = rollerSession.getAuthenticatedUser() != null
+                ? rollerSession.getAuthenticatedUser().getId() : "";
+        String salt = request.getParameter("salt");
+        if (salt == null) {
+            return false;
+        }
+
+        SaltCache saltCache = SaltCache.getInstance();
+        synchronized (saltCache) {
+            if (!Objects.equals(saltCache.get(salt), userId)) {
+                return false;
+            }
+            saltCache.remove(salt);
+        }
+        return true;
+    }
+
+    /**
+     * Returns true for a multipart form POST, which Struts parses after the
+     * servlet filters have run.
+     *
+     * @param request current request
+     * @return true for multipart/form-data POST requests
+     */
+    public static boolean isMultipartFormPost(HttpServletRequest request) {
+        if (!"POST".equalsIgnoreCase(request.getMethod())) {
+            return false;
+        }
+
+        String contentType = request.getContentType();
+        if (contentType == null) {
+            return false;
+        }
+
+        int parameterStart = contentType.indexOf(';');
+        String mediaType = parameterStart >= 0
+                ? contentType.substring(0, parameterStart) : contentType;
+        return MULTIPART_FORM_DATA.equals(mediaType.trim().toLowerCase(Locale.ENGLISH));
+    }
+}
--- a/app/src/main/java/org/apache/roller/weblogger/ui/core/filters/SaltValidator.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/core/filters/SaltValidator.java
@@ -22,6 +22,7 @@ package org.apache.roller.weblogger.ui.core.filters;
 import java.util.Locale;
 import java.util.Objects;
 
+import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 
 import org.apache.roller.weblogger.ui.core.RollerSession;
@@ -33,6 +34,9 @@ import org.apache.roller.weblogger.ui.rendering.util.cache.SaltCache;
 public final class SaltValidator {
 
     private static final String MULTIPART_FORM_DATA = "multipart/form-data";
+
+    public static final String VALIDATED_REQUEST_ATTRIBUTE =
+            SaltValidator.class.getName() + ".validated";
 
     private SaltValidator() {
     }
@@ -64,6 +68,19 @@ public final class SaltValidator {
             saltCache.remove(salt);
         }
         return true;
+    }
+
+    /**
+     * Validates and consumes the submitted salt or rejects the request.
+     *
+     * @param request current request
+     * @throws ServletException when the submitted salt is missing or invalid
+     */
+    public static void requireSubmittedSalt(HttpServletRequest request)
+            throws ServletException {
+        if (!consumeSubmittedSalt(request)) {
+            throw new ServletException("Security Violation");
+        }
     }
 
     /**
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/apache/roller/weblogger/ui/struts2/util/ValidateSaltInterceptor.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private static final long serialVersionUID = 2446434402795510394L;
+    private static final Log log = LogFactory.getLog(ValidateSaltInterceptor.class);
+
+    @Override
+    public String intercept(ActionInvocation invocation) throws Exception {
+        ActionContext context = invocation.getInvocationContext();
+        HttpServletRequest request = (HttpServletRequest) context.get(HTTP_REQUEST);
+
+        if (SaltValidator.isMultipartFormPost(request)
```
mraible: This runs again on every chained invocation of the same request. `bookmarksImport!save` is multipart and has `<result name="success" type="chain">bookmarks</result>`, so the second pass finds the salt already removed from `SaltCache` and throws `Security Violation` after the OPML import has been flushed. Simplest fix: after a successful `consumeSubmittedSalt`, set a request attribute (e.g. `request.setAttribute("salt.validated", Boolean.TRUE)`) and skip validation when it's present. Good candidate for a `ValidateSaltInterceptorTest` case.

Related: when Struts can't parse the multipart body (over `struts.multipart.maxSize`), `MultiPartRequestWrapper` has no fields, `getParameter("salt")` is null and this throws a 500 instead of letting the `fileUpload` interceptor's size error reach the form as it does today. Checking `((MultiPartRequestWrapper) request).hasErrors()` first would keep that UX.

app/src/main/java/org/apache/roller/weblogger/ui/core/filters/ValidateSaltFilter.java
```diff
@@ -31,53 +28,35 @@
...
+            if (SaltValidator.isMultipartFormPost(httpReq) && isStrutsAction(httpReq)) {
+                // Struts makes multipart parameters available after its upload
+                // interceptor; ValidateSaltInterceptor handles these requests.
+                chain.doFilter(request, response);
+                return;
+            }
 
-                // Remove salt from cache after successful validation
-                saltCache.remove(salt);
+            if (!SaltValidator.consumeSubmittedSalt(httpReq)) {
```
mraible: `Comments.jsp:429` reads `#comments_salt` once and sends it on every `commentdata` AJAX POST. With the salt removed on first use, editing a second comment without reloading fails here with a 500, and the `$.ajax` call has no error handler, so the save just silently doesn't happen. `CommentDataServlet` could return a fresh salt in its JSON (the response passes through `LoadSaltFilter`, so `request.getAttribute("salt")` is available) and the JS update `#comments_salt` from it.

app/src/main/java/org/apache/roller/weblogger/ui/core/filters/SaltValidator.java
```diff
@@ -0,0 +1,91 @@
...
+        String userId = rollerSession.getAuthenticatedUser() != null
+                ? rollerSession.getAuthenticatedUser().getId() : "";
+        String salt = request.getParameter("salt");
+        if (salt == null) {
+            return false;
+        }
+
+        SaltCache saltCache = SaltCache.getInstance();
+        synchronized (saltCache) {
+            if (!Objects.equals(saltCache.get(salt), userId)) {
```
mraible, line 67: Now that this is the only path, `cache.salt.size` (5000) and `cache.salt.timeout` (3600s) are user-visible limits: `LoadSaltFilter` mints a salt on every `/roller-ui` request and on every tiles FORWARD, so entries get evicted quickly on a multi-user site, and a form left open for over an hour is rejected with the entry text lost. Not a blocker, but please call it out in the description; raising the defaults would soften it.

app/src/main/java/org/apache/roller/weblogger/ui/core/filters/ValidateSaltFilter.java
```diff
@@ -31,53 +28,35 @@
...
-                salt = salt != null ? salt : httpReq.getParameter("salt");
-                SaltCache saltCache = SaltCache.getInstance();
-                if (salt == null || !Objects.equals(saltCache.get(salt), userId)) {
-                    if (log.isDebugEnabled()) {
-                        log.debug("Valid salt value not found on POST to URL : " + httpReq.getServletPath());
-                    }
-                    throw new ServletException("Security Violation");
-                }
+        if ("POST".equalsIgnoreCase(httpReq.getMethod())) {
+            if (SaltValidator.isMultipartFormPost(httpReq) && isStrutsAction(httpReq)) {
```
mraible: Edge case: Struts only wraps requests whose full `Content-Type` matches its `MULTIPART_FORM_DATA_REGEX` (boundary ≤ 70 chars, `boundary` before `charset`). A multipart POST this check accepts but Struts doesn't wrap skips validation here and then fails in the interceptor because `getParameter("salt")` is null. Browsers never send those, so probably fine, just noting that deferring on media type alone isn't quite the same decision Struts makes.

app/src/test/java/org/apache/roller/weblogger/ui/core/filters/SaltConfigurationTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class SaltConfigurationTest {
+
+    @Test
+    public void testSubmittedSaltIsValidatedBeforeResponseSaltIsLoaded() throws Exception {
+        String webXml = Files.readString(Path.of("src/main/webapp/WEB-INF/web.xml"));
```
mraible: cwd-relative, so this passes only when run from `app/`; the other two tests in this class use `getResourceAsStream`, and surefire sets `project.build.directory` for this module.

app/src/main/resources/struts.xml
```diff
@@ -55,6 +57,7 @@
                 <!-- <interceptor-ref name="scopedModelDriven"/> -->
                 <!-- <interceptor-ref name="modelDriven"/> -->
                 <interceptor-ref name="fileUpload"/>
+                <interceptor-ref name="ValidateSaltInterceptor"/>
```
mraible: The comment in `ValidateSaltFilter` says multipart params only exist after `fileUpload`, but Struts wraps the request in `MultiPartRequestWrapper` in `StrutsPrepareAndExecuteFilter` before any interceptor runs; `fileUpload` only copies file items into action params. So this ref could sit right after `exception` (validate before anything else does work), and `SaltConfigurationTest:51` shouldn't pin it as adjacent to `fileUpload`.

app/src/main/java/org/apache/roller/weblogger/ui/core/filters/ValidateSaltFilter.java
```diff
@@ -31,53 +28,35 @@
...
 
-                // Remove salt from cache after successful validation
-                saltCache.remove(salt);
+            if (!SaltValidator.consumeSubmittedSalt(httpReq)) {
                 if (log.isDebugEnabled()) {
-                    log.debug("Salt used and invalidated: " + salt);
+                    log.debug("Valid salt value not found on POST to URL : "
+                            + httpReq.getServletPath());
                 }
+                throw new ServletException("Security Violation");
```
mraible: Nit: this block is duplicated in the interceptor; a `SaltValidator.requireSubmittedSalt(request)` that throws would keep both rejection paths in sync.

app/src/test/java/org/apache/roller/weblogger/ui/core/filters/ValidateSaltFilterTest.java
```diff
@@ -129,42 +117,117 @@ public void testDoFilterWithPostMethodAndNullRollerSession() throws Exception {
...
+
+            assertThrows(ServletException.class,
+                    () -> filter.doFilter(request, response, chain));
+
+            verify(chain, never()).doFilter(request, response);
+        }
+    }
+
+    @Test
+    public void testValidationRunsBeforeResponseSaltGeneration() throws Exception {
```
mraible: This builds the chain in validate-then-load order and then asserts that order, so it can't fail if `web.xml` is swapped back; `SaltConfigurationTest` already covers the ordering.

app/src/test/java/org/apache/roller/weblogger/ui/core/filters/SaltConfigurationTest.java
```diff
@@ -0,0 +1,85 @@
...
+        assertTrue(adjacentInterceptors.matcher(strutsXml).find(),
+                "ValidateSaltInterceptor must immediately follow the upload interceptor");
+    }
+
+    @Test
+    public void testConfigurableSaltBypassIsRemoved() throws Exception {
+        String properties = readResource(
+                "/org/apache/roller/weblogger/config/roller.properties");
+
+        assertFalse(properties.contains("salt.ignored.urls"));
```
mraible: Asserting the literal `salt.ignored.urls` is absent from `roller.properties` means a migration note like `# salt.ignored.urls is no longer supported` breaks the build. The behaviour is already covered by the filter no longer reading the property.

mraible (commented):
Reviewed. Please address the two workflow regressions noted in the review before merge.

app/src/main/java/org/apache/roller/weblogger/ui/core/filters/SaltValidator.java
```diff
@@ -0,0 +1,108 @@
...
+     * @return true when no Roller session is present or the submitted salt is valid
+     */
+    public static boolean consumeSubmittedSalt(HttpServletRequest request) {
+        RollerSession rollerSession = RollerSession.getRollerSession(request);
+        if (rollerSession == null) {
+            return true;
+        }
+
+        String userId = rollerSession.getAuthenticatedUser() != null
+                ? rollerSession.getAuthenticatedUser().getId() : "";
```
mbien: https://github.com/apache/roller/pull/165#discussion_r3940931010

mbien (approved):

PR #168, opened 2026-08-29
Move authoring UI inline JS handlers to data attributes

snoopdave:
Several authoring pages embed values in inline JavaScript string literals and
then write them into the page with jQuery `.html()`. This refactor moves those
values into data attributes and writes them through text APIs, so JSPs stop
hand-concatenating markup and data.

## What changed

- Move server/user-controlled values out of inline JavaScript string literals
  into double-quoted, HTML-escaped `data-*` attributes.
- Bind delegated event listeners that read values as text and assign them via
  `textContent`, `.text()`, `.val()`, or DOM constructors.
- Replace the affected `.html()` writes with a text API.
- Use double-quoted attributes for weblog-content values in the media views.
- Cover templates, bookmarks/folders, categories, media views/chooser/success,
  entry lists, and entry editing, including the category iterator and
  delete-modal cases.
- Render theme descriptions as text and make `ThemeDataServlet` return valid
  JSON.

## Tests

- Values containing apostrophes, quotes, angle brackets, backslashes, newlines,
  and `</script>` remain text through each path.
- Cross-user authoring cases confirm one author's values stay text in another
  author or administrator session.
- A source audit confirms no affected value remains in an inline literal, a
  single-quoted attribute, or a `.html()` sink.

### DIFF
--- a/app/src/main/java/org/apache/roller/weblogger/ui/struts2/ajax/ThemeDataServlet.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/struts2/ajax/ThemeDataServlet.java
@@ -17,6 +17,7 @@
  */
 package org.apache.roller.weblogger.ui.struts2.ajax;
 
+import org.apache.commons.text.StringEscapeUtils;
 import org.apache.roller.weblogger.WebloggerException;
 import org.apache.roller.weblogger.business.WebloggerFactory;
 import org.apache.roller.weblogger.business.themes.SharedTheme;
@@ -80,17 +81,21 @@ public class ThemeDataServlet extends HttpServlet {
         }
         for (Iterator<SharedTheme> it = themes.iterator(); it.hasNext();) {
             SharedTheme theme = it.next();
+            // Theme metadata comes from theme.xml, which an operator can edit
+            // or install; escape it so a quote or newline cannot break out of
+            // the string and produce malformed JSON.
             pw.print("    { \"id\" : \"");
-            pw.print(theme.getId());
+            pw.print(StringEscapeUtils.escapeJson(theme.getId()));
             pw.print("\", ");
             pw.print("\"name\" : \"");
-            pw.print(theme.getName());
+            pw.print(StringEscapeUtils.escapeJson(theme.getName()));
             pw.print("\", ");
             pw.print("\"description\" : \"");
-            pw.print(theme.getDescription());
+            pw.print(StringEscapeUtils.escapeJson(theme.getDescription()));
             pw.print("\", ");
             pw.print("\"previewPath\" : \"");
-            pw.print("/themes" + "/" + theme.getId() + "/" + theme.getPreviewImage().getPath());
+            pw.print(StringEscapeUtils.escapeJson(
+                    "/themes" + "/" + theme.getId() + "/" + theme.getPreviewImage().getPath()));
             pw.print("\" }");
             if (it.hasNext()) {
                 pw.println(", ");
--- a/app/src/main/webapp/WEB-INF/jsps/editor/Bookmarks.jsp
+++ b/app/src/main/webapp/WEB-INF/jsps/editor/Bookmarks.jsp
@@ -706,8 +706,7 @@ We used to call them Bookmarks and Folders, now we call them Blogroll links and 
     }
 
 
-    // Author-supplied values travel in data attributes and are read back as
-    // text rather than being interpolated into inline handlers.
+    // Values come from data-* attributes and are bound via delegated listeners.
     $(document).on('click', '.bookmark-edit-link', function (event) {
         event.preventDefault();
         editBookmark($(this).attr('data-bookmark-id'),
--- a/app/src/main/webapp/WEB-INF/jsps/editor/Bookmarks.jsp
+++ b/app/src/main/webapp/WEB-INF/jsps/editor/Bookmarks.jsp
@@ -143,13 +143,13 @@ We used to call them Bookmarks and Folders, now we call them Blogroll links and 
 
                     <td align="center">
 
-                        <a href="#" onclick="editBookmark(
-                                '<s:property value="#bookmark.id"/>',
-                                '<s:property value="#bookmark.name"/>',
-                                '<s:property value="#bookmark.url"/>',
-                                '<s:property value="#bookmark.feedUrl"/>',
-                                '<s:property value="#bookmark.description"/>',
-                                '<s:property value="#bookmark.image"/>' )">
+                        <a href="#" class="bookmark-edit-link"
+                           data-bookmark-id="<s:property value="#bookmark.id"/>"
+                           data-bookmark-name="<s:property value="#bookmark.name"/>"
+                           data-bookmark-url="<s:property value="#bookmark.url"/>"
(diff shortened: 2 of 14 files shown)


app/src/main/webapp/WEB-INF/jsps/editor/Bookmarks.jsp
```diff
@@ -338,7 +338,7 @@ We used to call them Bookmarks and Folders, now we call them Blogroll links and
 
     function confirmDeleteFolder() {
         $('#boomarks_delete_folder_folderId').val($('#bookmarks_folderId:first').val());
-        $('#deleteBlogrollName').html('<s:property value="%{folder.name}"/>');
+        $('#deleteBlogrollName').text('<s:property value="%{folder.name}"/>');
```
mraible: `<s:property>` HTML-escapes with `escapeHtml4`, which doesn't touch `'`, so a folder named `Matt's Links` renders as `.text('Matt's Links')` and the SyntaxError takes out every function in this `<script>` block (delete/rename buttons stop working). And now that it's `.text()`, `Tom & Jerry` displays as `Tom &amp; Jerry`. Use `<s:property value="%{folder.name}" escapeJavaScript="true" escapeHtml="false"/>` here, or read the name from a `data-` attribute on the trigger element instead of inlining it in JS.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/ajax/ThemeDataServlet.java
```diff
@@ -80,17 +81,21 @@ public void doGet(
...
-            pw.print(theme.getId());
+            pw.print(StringEscapeUtils.escapeJson(theme.getId()));
             pw.print("\", ");
             pw.print("\"name\" : \"");
-            pw.print(theme.getName());
+            pw.print(StringEscapeUtils.escapeJson(theme.getName()));
             pw.print("\", ");
             pw.print("\"description\" : \"");
-            pw.print(theme.getDescription());
+            pw.print(StringEscapeUtils.escapeJson(theme.getDescription()));
```
mraible: `CreateWeblog.jsp:118` consumes this same field with `$('#themedescription').html(data.description)`, so a theme description is inert on Theme Edit but still rendered as markup on Create Weblog. Worth switching that call to `.text()` in this PR, since the audit test only scans `jsps/editor` and won't notice it.

app/src/main/webapp/WEB-INF/jsps/editor/Categories.jsp
```diff
@@ -288,7 +288,7 @@
     function showCategoryDeleteModal( id, name, inUse ) {
         $('#categoryRemove_removeId').val(id);
         $('#categoryEdit_bean_name').val(name);
-        $('#category-name').html(name);
+        $('#category-name').text(name);
         if ( inUse ) {
             $('#category-in-use').css('display','block');
```
mraible, line 293: Pre-existing, but since this function is being reworked: lines 294/297 toggle `#category-emtpy` while the element is `#category-empty` (line 266), so the "no entries in this category" message never shows.

app/src/main/webapp/WEB-INF/jsps/editor/ThemeEdit.jsp
```diff
@@ -246,7 +246,7 @@
         $.ajax({
             url: "<s:url value='themedata'/>",
             data: {theme: themeId}, success: function (data) {
-                $('#themeDescription').html(data.description);
+                $('#themeDescription').text(data.description);
```
mraible, line 249: Behaviour change worth a line in the description: a shared theme whose `theme.xml` `<description>` carries markup now renders it as literal text. The bundled themes are plain text, so I think that's fine, just disclose it.

app/src/test/java/org/apache/roller/weblogger/ui/struts2/editor/AuthoringUiSinkAuditTest.java
```diff
@@ -0,0 +1,163 @@
...
+ * This test enforces the two structural rules that keep those values inert: they
+ * travel in double-quoted <code>data-*</code> attributes rather than inline
+ * handler literals, and they are written to the DOM through a text API. It is a
+ * source audit rather than a behavioural test because the guarantee is a
+ * property of the whole page family, not of any one code path.
+ */
+public class AuthoringUiSinkAuditTest {
+
+    private static final Path EDITOR_JSP_DIR =
+            Paths.get("src", "main", "webapp", "WEB-INF", "jsps", "editor");
```
mraible: Two things here. This is cwd-relative, so it only runs from `app/`; surefire already sets `project.build.directory` for this module (see `ApplicationResourcesTest`), so deriving the path from that (or `basedir`) lets it run from an IDE rooted at the repo.

More importantly, the patterns give false assurance: `SCRIPT_VAR_LITERAL` only matches `var x = '<s:property`, `HANDLER_LITERAL` only matches a double-quoted handler attribute with a single-quoted inner literal, and `HTML_WRITE` ignores `.append`/`innerHTML`/`.attr('src')`. A single-quoted `<s:property>` as a call argument (`.text('<s:property .../>')`, `previewImage('<s:property .../>')`) or a single-quoted attribute (`href='<s:property .../>'`) isn't caught, and both remaining `Bookmarks.jsp` sinks pass this test green.

app/src/main/webapp/WEB-INF/jsps/editor/Categories.jsp
```diff
@@ -319,4 +320,30 @@
...
+        showCategoryDeleteModal($(this).attr('data-category-id'),
+                $(this).attr('data-category-name'),
+                $(this).attr('data-category-in-use') === 'true');
+    });
+
 </script>
+
+<%-- Source data for the "move entries to" select, carried as escaped
+     attributes rather than generated JavaScript literals. --%>
+<div id="category-option-data" style="display:none">
```
mraible: Nit: the `.category-edit-link` anchors above already carry `data-category-id` / `data-category-name` for every category, so `populateCategorySelect` could read those instead of maintaining a second serialised copy.

app/src/test/java/org/apache/roller/weblogger/ui/struts2/editor/AuthoringUiSinkAuditTest.java
```diff
@@ -0,0 +1,163 @@
...
+                + dir.toAbsolutePath() + " (run from the app module)");
+        try (Stream<Path> files = Files.list(dir)) {
+            return files.filter(p -> p.getFileName().toString().endsWith(".jsp"))
+                    .filter(p -> !EXCLUDED_FILES.contains(p.getFileName().toString()))
+                    .sorted()
+                    .collect(Collectors.toList());
+        }
+    }
+
+    private static String flatten(String source) {
```
mraible: Nit: `HANDLER_LITERAL` uses `\s*` and `[^"]*`, both of which already span newlines, so `flatten()` / `flattenSource` don't change any match and can go.

mraible (commented):
Reviewed. The output-handling changes are moving in the right direction. Please address the remaining consistency and test-coverage notes before merge.

mbien (approved):
looks fine

PR #169, opened 2026-08-29
Normalize frontpage directory parameters before rendering the bundled theme

snoopdave:
The bundled frontpage theme's blog-directory page filters its listing by a
`letter` request parameter and builds a back-link from a `weblog` parameter.
This change validates and normalizes both before use and escapes them at output.

## What changed

- Accept a single ASCII letter, normalize lowercase to uppercase, and require it
  to be a key in the business layer's weblog-letter map.
- Fall back to the full directory ("All weblogs") for missing or invalid input,
  without redirecting, erroring, or echoing the rejected value.
- Escape the accepted value at HTML output even after validation.
- Resolve the `weblog` parameter to an existing weblog and build the back-link
  from the resolved handle, escaped at output.

## Tests

- Missing input renders "All weblogs"; uppercase and lowercase valid keys select
  the same group.
- Multi-character, numeric, punctuation, and non-ASCII values fall back to the
  full listing and do not appear in the response, raw or encoded.
- Paging retains only the normalized, validated key.
- Strengthened business-map test asserts the exact A–Z key set.

### DIFF
--- a/app/src/main/webapp/themes/frontpage/_blogdirectory.vm
+++ b/app/src/main/webapp/themes/frontpage/_blogdirectory.vm
@@ -3,8 +3,8 @@
     ## Accept only a known A-Z key; otherwise render the full listing, exactly
     ## as a missing parameter does.
     #set($requestedLetter = $model.getRequestParameter("letter"))
-    #if($requestedLetter && $requestedLetter.length() == 1)
-        #set($candidateLetter = $requestedLetter.toUpperCase())
+    #if($requestedLetter)
+        #set($candidateLetter = $utils.toUpperCase($requestedLetter))
         #if($weblogLetterMap.containsKey($candidateLetter))
             #set($chosenLetter = $candidateLetter)
         #end
--- a/app/src/main/webapp/themes/frontpage/_blogdirectory.vm
+++ b/app/src/main/webapp/themes/frontpage/_blogdirectory.vm
@@ -1,8 +1,14 @@
-#if($model.getRequestParameter("letter"))
-        #set($chosenLetter = $model.getRequestParameter("letter"))
-    #end
+#set($weblogLetterMap = $site.getWeblogHandleLetterMap())
 
-    #set($weblogLetterMap = $site.getWeblogHandleLetterMap())
+    ## Accept only a known A-Z key; otherwise render the full listing, exactly
+    ## as a missing parameter does.
+    #set($requestedLetter = $model.getRequestParameter("letter"))
+    #if($requestedLetter && $requestedLetter.length() == 1)
+        #set($candidateLetter = $requestedLetter.toUpperCase())
+        #if($weblogLetterMap.containsKey($candidateLetter))
+            #set($chosenLetter = $candidateLetter)
+        #end
+    #end
     <div class="letterMap">
     <p>
     #set($firstLetterDone = 0)
@@ -22,7 +28,7 @@
     </div>
 
     #if($chosenLetter)
-       <h2 class="pageTitle">Weblogs starting with $chosenLetter</h2>
+       <h2 class="pageTitle">Weblogs starting with $utils.escapeHTML($chosenLetter)</h2>
     #else
        <h2 class="pageTitle">All weblogs</h2>
     #end
--- a/app/src/main/webapp/themes/frontpage/directory.vm
+++ b/app/src/main/webapp/themes/frontpage/directory.vm
@@ -31,10 +31,15 @@
 
             <div id="tabContent">
                 <div id="directory">
-                #if($model.getRequestParameter("weblog"))
-                    #set($handle = $model.getRequestParameter("weblog"))
-                    <a href="?letter=$utils.left($handle,1)">Back to blog directory</a>
-                    #set($profileWeblog = $site.getWeblog($handle))
+                ## Render the profile only for a weblog that exists, and build
+                ## the back-link from the resolved weblog's own handle.
+                #set($profileWeblog = false)
+                #set($requestedHandle = $model.getRequestParameter("weblog"))
+                #if($requestedHandle)
+                    #set($profileWeblog = $site.getWeblog($requestedHandle))
+                #end
+                #if($profileWeblog)
+                    <a href="?letter=$utils.escapeHTML($utils.left($profileWeblog.handle,1))">Back to blog directory</a>
                     #includeTemplate($model.weblog "_blogprofile")
                 #else                 
                    #set($pageLength = $maxResults)
--- a/app/src/main/webapp/themes/frontpage/directory.vm
+++ b/app/src/main/webapp/themes/frontpage/directory.vm
@@ -35,11 +35,16 @@
                 ## the back-link from the resolved weblog's own handle.
                 #set($profileWeblog = false)
                 #set($requestedHandle = $model.getRequestParameter("weblog"))
-                #if($requestedHandle)
+                #if($requestedHandle && $requestedHandle.matches("[A-Za-z0-9_]+"))
                     #set($profileWeblog = $site.getWeblog($requestedHandle))
                 #end
                 #if($profileWeblog)
-                    <a href="?letter=$utils.escapeHTML($utils.left($profileWeblog.handle,1))">Back to blog directory</a>
+                    #set($resolvedFirst = $utils.left($profileWeblog.handle,1))
+                    #if($resolvedFirst.matches("[A-Za-z]"))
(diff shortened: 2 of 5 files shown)


app/src/main/webapp/themes/frontpage/directory.vm
```diff
@@ -31,10 +31,15 @@
...
-                #if($model.getRequestParameter("weblog"))
-                    #set($handle = $model.getRequestParameter("weblog"))
-                    <a href="?letter=$utils.left($handle,1)">Back to blog directory</a>
-                    #set($profileWeblog = $site.getWeblog($handle))
+                ## Render the profile only for a weblog that exists, and build
+                ## the back-link from the resolved weblog's own handle.
+                #set($profileWeblog = false)
+                #set($requestedHandle = $model.getRequestParameter("weblog"))
+                #if($requestedHandle)
+                    #set($profileWeblog = $site.getWeblog($requestedHandle))
```
mraible, line 39: `$site.getWeblog()` still receives the raw parameter. `JPAWeblogManagerImpl.getWeblogByHandle` throws `WebloggerException("Invalid handle: '...'")` for anything outside `[A-Za-z0-9_]`, and `SiteModel.getWeblog` logs that at ERROR with a stack trace. A cheap pre-check such as `#if($requestedHandle && $requestedHandle.matches("[A-Za-z0-9_]+"))` keeps garbage from reaching the manager.

app/src/main/webapp/themes/frontpage/directory.vm
```diff
@@ -31,10 +31,15 @@
...
+                #end
+                #if($profileWeblog)
```
mraible, line 41: Handles may start with a digit (`username.allowedChars` defaults to `A-Za-z0-9`), and `getWeblogsByLetter` handled `?letter=2` before. Now `2` isn't a key in the A-Z map, so "Back to blog directory" from such a profile lands on the unfiltered list. Either accept the resolved handle's first character in `_blogdirectory.vm` (it's trusted) or omit the `letter` param when it isn't A-Z.

app/src/main/webapp/themes/frontpage/_blogdirectory.vm
```diff
@@ -1,8 +1,14 @@
...
-        #set($chosenLetter = $model.getRequestParameter("letter"))
-    #end
+#set($weblogLetterMap = $site.getWeblogHandleLetterMap())
 
-    #set($weblogLetterMap = $site.getWeblogHandleLetterMap())
+    ## Accept only a known A-Z key; otherwise render the full listing, exactly
+    ## as a missing parameter does.
+    #set($requestedLetter = $model.getRequestParameter("letter"))
+    #if($requestedLetter && $requestedLetter.length() == 1)
+        #set($candidateLetter = $requestedLetter.toUpperCase())
```
mraible: Minor: `toUpperCase()` uses the JVM default locale, so on a Turkish-locale server `i` becomes `İ` and `?letter=i` is rejected while `?letter=I` works. A `Locale.ROOT` upper-case helper on `$utils` would avoid it.

app/src/main/webapp/themes/frontpage/_blogdirectory.vm
```diff
@@ -1,8 +1,14 @@
-#if($model.getRequestParameter("letter"))
-        #set($chosenLetter = $model.getRequestParameter("letter"))
-    #end
+#set($weblogLetterMap = $site.getWeblogHandleLetterMap())
 
-    #set($weblogLetterMap = $site.getWeblogHandleLetterMap())
+    ## Accept only a known A-Z key; otherwise render the full listing, exactly
+    ## as a missing parameter does.
+    #set($requestedLetter = $model.getRequestParameter("letter"))
+    #if($requestedLetter && $requestedLetter.length() == 1)
```
mraible: Nit: the `length() == 1` check and `$candidateLetter` collapse to `#if($requestedLetter && $weblogLetterMap.containsKey($requestedLetter.toUpperCase()))`, since every key is one character. Keep the `null` guard, `TreeMap.containsKey(null)` throws.

app/src/main/webapp/themes/frontpage/directory.vm
```diff
@@ -31,10 +31,15 @@
...
+                    <a href="?letter=$utils.escapeHTML($utils.left($profileWeblog.handle,1))">Back to blog directory</a>
```
mraible: Nit: `escapeHTML` can't change anything here, `$profileWeblog.handle` already passed the `[A-Za-z0-9_]` check in `getWeblogByHandle`.

app/src/main/webapp/themes/frontpage/directory.vm
```diff
@@ -31,10 +31,15 @@
 
             <div id="tabContent">
                 <div id="directory">
-                #if($model.getRequestParameter("weblog"))
-                    #set($handle = $model.getRequestParameter("weblog"))
-                    <a href="?letter=$utils.left($handle,1)">Back to blog directory</a>
-                    #set($profileWeblog = $site.getWeblog($handle))
+                ## Render the profile only for a weblog that exists, and build
+                ## the back-link from the resolved weblog's own handle.
+                #set($profileWeblog = false)
```
mraible, line 36: Nit: nothing sets `$profileWeblog` before this and Velocity 2.4 assigns `null` from `#set`, so the initialiser is a no-op.

app/src/test/java/org/apache/roller/weblogger/ui/rendering/velocity/FrontpageDirectoryRenderingTest.java
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * The sibling directory template resolves a weblog handle from the query
+     * string. It cannot be rendered standalone here because it pulls in other
+     * templates through #includeTemplate, so this is a structural check: the
+     * link must be built from the resolved weblog rather than the raw
+     * parameter, and escaped at output.
+     */
+    @Test
+    public void directoryTemplateValidatesTheWeblogParameter() throws Exception {
```
mraible, line 206: This asserts source substrings of `directory.vm`, so a whitespace change breaks it while a re-introduced raw parameter under another variable name passes. `#includeTemplate` is a plain velocimacro in `WEB-INF/velocity/weblog.vm`, so the test can stub it inline (`#macro(includeTemplate $w $p)#end` before `#parse('directory.vm')`) and assert the rendered output: no profile for an unknown handle, back-link built from the resolved handle for a known one.

app/src/test/java/org/apache/roller/weblogger/ui/rendering/velocity/FrontpageDirectoryRenderingTest.java
```diff
@@ -0,0 +1,209 @@
...
+
+    private static VelocityEngine engine;
+
+    @BeforeAll
+    public static void setUpEngine() {
+        Properties props = new Properties();
+        props.setProperty("resource.loaders", "file");
+        props.setProperty("resource.loader.file.class",
+                "org.apache.velocity.runtime.resource.loader.FileResourceLoader");
+        props.setProperty("resource.loader.file.path", THEME_DIR);
```
mraible, line 60: `app/pom.xml` already copies `src/main/webapp/themes/**` onto the test classpath (that's how `themes.dir` in `roller-custom.properties` works), so a `ClasspathResourceLoader` rooted at `themes/frontpage/` drops the working-directory dependence.

app/src/test/java/org/apache/roller/weblogger/ui/rendering/velocity/FrontpageDirectoryRenderingTest.java
```diff
@@ -0,0 +1,209 @@
...
+                map.put(String.valueOf(c), 1L);
+            }
+            return map;
+        }
+        public StubPager getWeblogsByLetterPager(String letter, int offset, int length) {
+            return new StubPager();
+        }
+    }
+
+    public static class StubUtils {
```
mraible: Nit: `UtilitiesModel` has a no-arg constructor and `left` / `escapeHTML` are stateless, so `ctx.put("utils", new UtilitiesModel())` tests the real `escapeHtml4` instead of a four-replace stand-in.

mraible (commented):
The letter gating and building the back-link from the resolved handle are good. Two things keep the `weblog` parameter from being "validated before use" as the title says (inline), plus test and cleanup nits. Happy to approve once the first two are addressed.

mbien (approved):

PR #170, opened 2026-08-29
Move frontpage selection into the administrator setup workflow

snoopdave:
The setup page is reachable without a login so a freshly installed site can be
bootstrapped before any user exists. This change makes that page a read-only
bootstrap view and moves the frontpage-weblog selection into a separate
global-administrator action.

## What changed

- While the site has zero users, show only a minimal public bootstrap page —
  no weblog lists, counts, or mutating action.
- Make the initial frontpage selection a separate POST action requiring global
  administrator permission, allowed only while no frontpage weblog has been set.
- Route later changes through `GlobalConfig`.
- Centralize canonical-handle resolution and enabled-weblog validation in one
  service used by both write paths; reject blank, missing, nonexistent, and
  disabled weblogs and treat a missing aggregation checkbox as false.
- Persist both properties together and invalidate local site-wide page/feed
  caches, including the site-wide last-modified value.
- Fail closed when `users.firstUserAdmin=false` until a global administrator is
  provisioned.

## Tests

- Anonymous zero-user page; anonymous and non-admin access after users exist;
  disabled admin and global-admin callers.
- Both first-user-admin settings; initial, concurrent, and already-configured
  writes; the retired `setup!save` action no longer resolves.
- Invalid weblog handles, null aggregation, successful reads, and local cache
  invalidation through both write paths.

### DIFF
--- a/app/src/main/java/org/apache/roller/weblogger/business/FrontpageSettings.java
+++ b/app/src/main/java/org/apache/roller/weblogger/business/FrontpageSettings.java
@@ -1,0 +1,139 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  The ASF licenses this file to You
+ * under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied.  See the License for the specific language governing
+ * permissions and limitations under the License.  For additional
+ * information regarding copyright in this work, please see the NOTICE
+ * file in the top level directory of this distribution.
+ */
+package org.apache.roller.weblogger.business;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.roller.weblogger.WebloggerException;
+import org.apache.roller.weblogger.pojos.RuntimeConfigProperty;
+import org.apache.roller.weblogger.pojos.Weblog;
+import org.apache.roller.weblogger.ui.rendering.util.cache.SiteWideCache;
+import org.apache.roller.weblogger.ui.rendering.util.cache.WeblogFeedCache;
+import org.apache.roller.weblogger.ui.rendering.util.cache.WeblogPageCache;
+
+/**
+ * Reads and writes the site frontpage weblog settings.
+ *
+ * <p>Two screens change these values: the one-time setup screen used to choose
+ * a frontpage while the site is being installed, and the global configuration
+ * screen used afterwards. Both go through here so that the handle is resolved
+ * and validated the same way, both properties move together, and the rendered
+ * page and feed caches are invalidated consistently.
+ */
+public final class FrontpageSettings {
+
+    public static final String HANDLE_PROPERTY = "site.frontpage.weblog.handle";
+    public static final String AGGREGATED_PROPERTY = "site.frontpage.weblog.aggregated";
+
+    private FrontpageSettings() {
+    }
+
+    /**
+     * Resolves a submitted handle to a weblog that actually exists and is
+     * enabled.
+     *
+     * @return the weblog, or null when the handle is blank, unknown or refers
+     *         to a disabled weblog
+     */
+    public static Weblog resolveWeblog(String handle) throws WebloggerException {
+        if (StringUtils.isBlank(handle)) {
+            return null;
+        }
+        return WebloggerFactory.getWeblogger().getWeblogManager()
+                .getWeblogByHandle(handle.trim(), Boolean.TRUE);
+    }
+
+    /** @return the configured frontpage handle, or null when none is set. */
+    public static String getConfiguredHandle() throws WebloggerException {
+        RuntimeConfigProperty prop = WebloggerFactory.getWeblogger()
+                .getPropertiesManager().getProperty(HANDLE_PROPERTY);
+        if (prop == null || StringUtils.isBlank(prop.getValue())) {
+            return null;
+        }
+        return prop.getValue();
+    }
+
+    /** @return true when a frontpage weblog has already been chosen. */
+    public static boolean isConfigured() throws WebloggerException {
+        return getConfiguredHandle() != null;
+    }
+
+    /**
+     * Validates and stores the frontpage selection.
+     *
+     * <p>Both properties are written before the single flush so the pair cannot
+     * be left half-applied, and the canonical handle from the resolved weblog is
+     * stored rather than the submitted text. A missing aggregation value is
+     * treated as false, which is what an unchecked checkbox means.
+     *
+     * @param handle     submitted weblog handle
+     * @param aggregated submitted aggregation flag; null means false
+     * @throws InvalidFrontpageWeblogException when the handle does not name an
+     *         existing, enabled weblog
+     */
+    public static void apply(String handle, Boolean aggregated)
+            throws WebloggerException {
+
+        Weblog weblog = resolveWeblog(handle);
+        if (weblog == null) {
+            throw new InvalidFrontpageWeblogException(handle);
+        }
+
+        PropertiesManager mgr = WebloggerFactory.getWeblogger().getPropertiesManager();
+
+        RuntimeConfigProperty handleProp = mgr.getProperty(HANDLE_PROPERTY);
+        handleProp.setValue(weblog.getHandle());
+        mgr.saveProperty(handleProp);
+
+        RuntimeConfigProperty aggregatedProp = mgr.getProperty(AGGREGATED_PROPERTY);
+        aggregatedProp.setValue(Boolean.toString(Boolean.TRUE.equals(aggregated)));
+        mgr.saveProperty(aggregatedProp);
+
+        WebloggerFactory.getWeblogger().flush();
+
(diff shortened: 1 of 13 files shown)


mraible (commented):
Reviewed. The setup-flow changes are moving in the right direction. Please resolve the front-page and setup regressions noted in the review before merge.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/core/Setup.java
```diff
@@ -64,46 +70,49 @@ public boolean isWeblogRequired() {
...
+        // thing that can usefully be done.
+        if (getUserCount() == 0) {
+            setBootstrap(true);
+            return SUCCESS;
+        }
 
-            WebloggerFactory.getWeblogger().flush();
+        // Beyond that point this is a site configuration screen.
+        if (!isUserIsAdmin()) {
+            return DENIED;
```
mraible: `index.jsp` (lines 29-31) forwards `/` here whenever the frontpage handle is blank, so returning `DENIED` for non-admins makes the home page an access-denied tile for every visitor of a site that has users but no frontpage yet (`checkPermission` with a `null` user throws and `isUserIsAdmin()` swallows it to false). The page only showed counts and links before; I'd keep `execute()` open and gate the form in the JSP on `isUserIsAdmin()`, leaving the POST admin-only.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/core/Setup.java
```diff
@@ -64,46 +70,49 @@ public boolean isWeblogRequired() {
...
+        // contents is exposed here: registering the first user is the only
+        // thing that can usefully be done.
+        if (getUserCount() == 0) {
+            setBootstrap(true);
+            return SUCCESS;
+        }
 
-            WebloggerFactory.getWeblogger().flush();
+        // Beyond that point this is a site configuration screen.
+        if (!isUserIsAdmin()) {
```
mraible: With `users.firstUserAdmin=false` (documented in `roller.properties`) no account ever passes `isUserIsAdmin()`, and `frontpageSetup!save` and `globalConfig` are admin-only too, so the frontpage can never be chosen and `/` stays broken. The removed `setup!save` was the only non-admin path; if it goes, the description should say how such installs are expected to finish setup.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/core/Setup.java
```diff
@@ -64,46 +70,49 @@ public boolean isWeblogRequired() {
...
+            return DENIED;
+        }
 
-            addMessage("frontpageConfig.values.saved");
+        try {
+            if (FrontpageSettings.isConfigured()) {
```
mraible: `JPAWeblogManagerImpl.removeWeblog` doesn't clear `site.frontpage.weblog.handle`, so after the frontpage weblog is deleted `isConfigured()` is still true and this redirects to `/`, which forwards to the missing weblog. On master the admin could come back here and pick another one. Either treat a handle that doesn't resolve as unconfigured, or clear the property in `removeWeblog`.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/core/FrontpageSetup.java
```diff
@@ -0,0 +1,122 @@
...
+                addError("frontpageConfig.alreadyConfigured");
+                return "home";
+            }
+
+            FrontpageSettings.apply(frontpageBlog, aggregated);
+            addMessage("frontpageConfig.values.saved");
+
+        } catch (FrontpageSettings.InvalidFrontpageWeblogException ex) {
+            addError("frontpageConfig.invalidWeblog");
+            return INPUT;
```
mraible, line 100: `struts.xml` maps `input` for this action to the `.Setup` tile, but `Setup.jsp` reads `userCount`, `blogCount`, `weblogs` and `bootstrap`, none of which exist on `FrontpageSetup`, so every `<s:if>` is false and the admin sees the error above three empty panels with no chooser. Either redirect back to `setup` carrying the message, or give this action the same properties.

app/src/main/webapp/WEB-INF/jsps/core/Setup.jsp
```diff
@@ -93,7 +93,8 @@
 
         <s:if test="blogCount > 0">
```
mraible: The heading was changed to `blogCount > 0 && !bootstrap` but this guard wasn't, so in bootstrap mode (weblogs exist, zero users, `weblogs` is `null`) anonymous visitors still get the form with an empty select and a Save that POSTs to an admin-only action.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/core/FrontpageSetup.java
```diff
@@ -0,0 +1,122 @@
...
+
+        try {
+            // Re-read immediately before writing so that a second submission
+            // arriving alongside the first cannot replace the winner. This
+            // narrows the window rather than closing it outright; the two
+            // submissions would have to interleave within this method, and the
+            // losing caller is told the choice is already made.
+            if (FrontpageSettings.isConfigured()) {
+                addError("frontpageConfig.alreadyConfigured");
+                return "home";
```
mraible: `rollerStack` has no `MessageStore` interceptor, so this `addError` is discarded by the `redirectAction`; the second admin just lands on the winner's frontpage. The comment above says the loser is told, but they aren't.

app/src/main/java/org/apache/roller/weblogger/business/FrontpageSettings.java
```diff
@@ -0,0 +1,139 @@
...
+     *
+     * @return the weblog, or null when the handle is blank, unknown or refers
+     *         to a disabled weblog
+     */
+    public static Weblog resolveWeblog(String handle) throws WebloggerException {
+        if (StringUtils.isBlank(handle)) {
+            return null;
+        }
+        return WebloggerFactory.getWeblogger().getWeblogManager()
+                .getWeblogByHandle(handle.trim(), Boolean.TRUE);
```
mraible, line 53: `getWeblogByHandle` throws `WebloggerException("Invalid handle")` for anything outside `[A-Za-z0-9_]`, so a POST with `frontpageBlog=my-blog` takes the generic `WebloggerException` branch (stack trace at ERROR, "Error saving properties") instead of the invalid-weblog message this method documents. Pre-check the handle or catch that case here.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/admin/GlobalConfig.java
```diff
@@ -209,6 +210,22 @@ public String save() {
...
+            } else if ( FrontpageSettings.HANDLE_PROPERTY.equals(propertyDef.getName())
+                    && incomingProp != null ) {
+                // Declared as a plain string, but it names a weblog, so it
+                // is resolved through the same service as the setup path. The
+                // stored value is always a weblog that exists and is enabled.
+                try {
+                    if (FrontpageSettings.resolveWeblog(incomingProp) == null) {
+                        addError("frontpageConfig.invalidWeblog");
+                    } else {
+                        updProp.setValue( incomingProp.trim() );
```
mraible: This validates via `resolveWeblog` but then stores the raw value itself instead of calling `FrontpageSettings.apply()`, so changing the frontpage from Global Config skips the cache invalidation the class javadoc promises for "both screens" and keeps serving the old weblog's cached pages and feeds.

app/src/main/java/org/apache/roller/weblogger/business/FrontpageSettings.java
```diff
@@ -0,0 +1,139 @@
...
+ * information regarding copyright in this work, please see the NOTICE
+ * file in the top level directory of this distribution.
+ */
+package org.apache.roller.weblogger.business;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.roller.weblogger.WebloggerException;
+import org.apache.roller.weblogger.pojos.RuntimeConfigProperty;
+import org.apache.roller.weblogger.pojos.Weblog;
+import org.apache.roller.weblogger.ui.rendering.util.cache.SiteWideCache;
```
mraible: First `business` → `ui.rendering` dependency in the tree; `SiteWideCache` / `WeblogPageCache` / `WeblogFeedCache` are rendering singletons that bootstrap caches on `getInstance()`. The invalidation belongs in the actions (`FrontpageSetup` / `GlobalConfig`), with `FrontpageSettings` just resolving and storing.

app/src/test/java/org/apache/roller/weblogger/ui/struts2/core/FrontpageSetupAccessTest.java
```diff
@@ -0,0 +1,150 @@
...
+    public void publicSetupPageStillNeedsNoUserButCannotWrite() throws IOException {
+        Setup setup = new Setup();
+        assertFalse(setup.isUserRequired(),
+                "the bootstrap page must stay reachable on a site with no users");
+
+        String struts = read(STRUTS_XML);
+        int setupIdx = struts.indexOf("name=\"setup\"");
+        assertTrue(setupIdx > 0, "setup action not found in struts.xml");
+        String setupBlock = struts.substring(setupIdx, struts.indexOf("</action>", setupIdx));
+        assertFalse(setupBlock.contains("save"),
```
mraible: These assert on source text (`contains("save")`, `method=\"post\"`), so an XML comment containing `save` in the setup block fails the build while a reachable-but-unguarded action passes. `requiredGlobalPermissionActions` is the only behavioural check here; I'd drop the text ones.

mbien (approved):

PR #171, opened 2026-08-29
Disable vendor extension types on the XML-RPC servlet

snoopdave:
The XML-RPC servlet is configured to accept the library's non-standard vendor
extension types, which Roller does not use, and its mapped endpoint answers
requests even when the XML-RPC feature is switched off. This change turns the
extensions off and makes the disabled-by-default toggle close the endpoint.

## What changed

- Disable vendor extension types (`enabledForExtensions=false`, or drop the
  enabling servlet init-param), since Roller uses only the standard XML-RPC
  value types.
- Gate the mapped endpoint on `webservices.enableXmlRpc`, so a disabled API is
  closed at the endpoint instead of relying on per-handler checks.
- Preserve ordinary XML-RPC values and calls when the feature is enabled.

## Tests

- With the toggle off, both ordinary and extension-type requests are rejected at
  the endpoint.
- With the toggle on, an extension-type value is rejected while an ordinary
  XML-RPC call reaches its handler.

### DIFF
--- a/app/src/main/java/org/apache/roller/weblogger/ui/core/filters/XmlRpcEnabledFilter.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/core/filters/XmlRpcEnabledFilter.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  The ASF licenses this file to You
+ * under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.  For additional information regarding
+ * copyright in this work, please see the NOTICE file in the top level
+ * directory of this distribution.
+ */
+
+package org.apache.roller.weblogger.ui.core.filters;
+
+import java.io.IOException;
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletResponse;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.roller.weblogger.config.WebloggerRuntimeConfig;
+
+/**
+ * Gates the XML-RPC endpoint on <code>webservices.enableXmlRpc</code>, ahead of
+ * the servlet, so a disabled service is not reachable.
+ */
+public class XmlRpcEnabledFilter implements Filter {
+
+    private static final Log LOG = LogFactory.getLog(XmlRpcEnabledFilter.class);
+
+    @Override
+    public void init(FilterConfig filterConfig) {
+        // nothing to configure
+    }
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response,
+                         FilterChain chain) throws IOException, ServletException {
+
+        if (!WebloggerRuntimeConfig.getBooleanProperty("webservices.enableXmlRpc")) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("XML-RPC service is disabled; rejecting request");
+            }
+            ((HttpServletResponse) response).sendError(HttpServletResponse.SC_NOT_FOUND);
+            return;
+        }
+
+        chain.doFilter(request, response);
+    }
+
+    @Override
+    public void destroy() {
+        // nothing to release
+    }
+}
--- a/app/src/main/java/org/apache/roller/weblogger/ui/core/filters/XmlRpcEnabledFilter.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/core/filters/XmlRpcEnabledFilter.java
@@ -49,10 +49,11 @@ public class XmlRpcEnabledFilter implements Filter {
                          FilterChain chain) throws IOException, ServletException {
 
         if (!WebloggerRuntimeConfig.getBooleanProperty("webservices.enableXmlRpc")) {
-            if (LOG.isDebugEnabled()) {
-                LOG.debug("XML-RPC service is disabled; rejecting request");
-            }
-            ((HttpServletResponse) response).sendError(HttpServletResponse.SC_NOT_FOUND);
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/apache/roller/weblogger/ui/core/filters/XmlRpcEnabledFilter.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response,
+                         FilterChain chain) throws IOException, ServletException {
+
+        if (!WebloggerRuntimeConfig.getBooleanProperty("webservices.enableXmlRpc")) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("XML-RPC service is disabled; rejecting request");
+            }
+            ((HttpServletResponse) response).sendError(HttpServletResponse.SC_NOT_FOUND);
```
mraible: `sendError(404)` gets routed to `/roller-ui/errors/404.jsp`, so a disabled endpoint now answers a desktop client's POST with an HTML page instead of the XML-RPC fault `BLOGGERAPI_DISABLED_MSG` that `BaseAPIHandler` still produces (line 152) but can no longer reach. Clients like MarsEdit will show a generic "invalid response". Consider writing a small XML-RPC fault with `text/xml` here, or at least `setStatus(404)` with a plain-text body so the error JSP isn't rendered.

app/src/main/java/org/apache/roller/weblogger/ui/core/filters/XmlRpcEnabledFilter.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    public void init(FilterConfig filterConfig) {
+        // nothing to configure
+    }
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response,
+                         FilterChain chain) throws IOException, ServletException {
+
+        if (!WebloggerRuntimeConfig.getBooleanProperty("webservices.enableXmlRpc")) {
```
mraible, line 51: `getBooleanProperty` swallows exceptions and returns false, so a DB hiccup or a request before bootstrap answers 404 rather than an error. A `LOG.warn` when refusing would make that diagnosable; debug is easy to miss.

app/src/main/webapp/WEB-INF/web.xml
```diff
@@ -154,6 +159,11 @@
 	</filter-mapping>
 
     <!-- Request mapping, this is what allows the urls to work -->
+    <filter-mapping>
```
mraible, line 162: Nit: no `<dispatcher>` so this applies to REQUEST only; nothing forwards to `/roller-services/xmlrpc` today, but adding `FORWARD` / `INCLUDE` makes the gate match what the javadoc claims.

app/src/test/java/org/apache/roller/weblogger/webservices/xmlrpc/XmlRpcExtensionTypeTest.java
```diff
@@ -0,0 +1,103 @@
...
+        assertTrue(webXml.contains("XmlRpcEnabledFilter"),
+                "a filter must gate the XML-RPC endpoint");
+        int mapping = webXml.indexOf("<filter-name>XmlRpcEnabledFilter</filter-name>",
+                webXml.indexOf("<filter-mapping>"));
+        assertTrue(mapping > 0, "the gating filter must be mapped");
+        assertTrue(webXml.indexOf("/roller-services/xmlrpc", mapping) > 0,
+                "the gating filter must be mapped to the XML-RPC endpoint");
+    }
+
+    /** Ordinary XML-RPC calls must still parse with extensions disabled. */
```
mraible: The description says extension types are rejected and the filter is tested, but this class only parses an ordinary call. A request containing `<ex:serializable>` (or `<ex:nil/>`) asserted to fail with extensions disabled, plus an `XmlRpcEnabledFilter` test with a mocked `sendError`, would actually pin the security property this PR introduces.

app/src/test/java/org/apache/roller/weblogger/webservices/xmlrpc/XmlRpcExtensionTypeTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /** The endpoint is closed by a filter while the service is switched off. */
+    @Test
+    public void endpointIsGatedWhileTheServiceIsDisabled() throws Exception {
+        String webXml = new String(Files.readAllBytes(WEB_XML), StandardCharsets.UTF_8);
+        assertTrue(webXml.contains("XmlRpcEnabledFilter"),
+                "a filter must gate the XML-RPC endpoint");
+        int mapping = webXml.indexOf("<filter-name>XmlRpcEnabledFilter</filter-name>",
+                webXml.indexOf("<filter-mapping>"));
+        assertTrue(mapping > 0, "the gating filter must be mapped");
```
mraible: `indexOf("/roller-services/xmlrpc", mapping) > 0` is also satisfied by the later `<servlet-mapping>` for `XmlRpcServlet`, so a typo in the filter's `url-pattern` still passes. Parse the `<filter-mapping>` element and compare its `url-pattern` directly.

app/src/test/java/org/apache/roller/weblogger/webservices/xmlrpc/XmlRpcExtensionTypeTest.java
```diff
@@ -0,0 +1,103 @@
...
+ * Covers how the XML-RPC endpoint treats vendor extension types.
+ *
+ * <p>Roller's XML-RPC API uses only the standard value types, so the library's
+ * vendor extension types are switched off in <code>web.xml</code> and the
+ * endpoint is closed entirely while the service is disabled.
+ */
+public class XmlRpcExtensionTypeTest {
+
+    private static final Path WEB_XML =
+            Paths.get("src", "main", "webapp", "WEB-INF", "web.xml");
```
mraible: cwd-relative; surefire sets `project.build.directory` for this module and `ApplicationResourcesTest` / `SQLScriptRunnerTest` derive their paths from it.

app/src/test/java/org/apache/roller/weblogger/webservices/xmlrpc/XmlRpcExtensionTypeTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private String parseRequest(String xml, boolean extensionsEnabled) throws Exception {
+        XmlRpcServer server = new XmlRpcServer();
```
mraible, line 70: Nit: `extensionsEnabled` is only ever `false`, the `(XmlRpcController)` cast is a no-op, and `XMLReaderFactory` has been deprecated since Java 9 (`SAXParserFactory.newInstance().newSAXParser().getXMLReader()`).

mraible (approved):
Reviewed. The configuration change is reasonable. Please address the compatibility and test-coverage notes before merge, and document the remaining outbound-client behavior separately.

mbien (approved):
more convenient than removal of the XmlRpcServlet servlet section of your web.xml :)

PR #172, opened 2026-08-29
Resolve weblog template resources within the active theme

snoopdave:
Weblog template rendering currently loads resources from the application
classpath in addition to the theme namespace. This change confines template
resource resolution to the active theme and restricts include targets to the
shapes a template actually takes.

## What changed

- Drop the classpath resource loader from the loader set used for weblog
  rendering, and align the stale test copy of `velocity.properties`.
- Register an include event handler that confines `#include` / `#parse` to the
  theme namespace and refuses absolute paths, upward traversal, schemes, and
  names that are neither a stored template id nor a `.vm` file.
- Keep `SecureUberspector` unchanged.

## Tests

`ThemeIncludeConfinementTest` covers loader set, handler registration, and
uberspector configuration; handler unit cases for refused and allowed names; and
an end-to-end engine test that measures each control independently.

### DIFF
--- a/app/src/main/java/org/apache/roller/weblogger/ui/rendering/velocity/ThemeIncludeEventHandler.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/rendering/velocity/ThemeIncludeEventHandler.java
@@ -38,12 +38,6 @@ import org.apache.velocity.context.Context;
  * needs to leave the namespace, so a name that is absolute, walks upward, or
  * carries a scheme is refused.
  *
- * <p>Names are also held to the shapes a template can actually take: a stored
- * template id, which carries no extension, or a Velocity template file. A name
- * that asks for some other kind of file is not a template reference at all, and
- * refusing it keeps the directives pointed at templates no matter what a loader
- * further down happens to be able to resolve.
- *
  * <p>Returning null tells Velocity not to resolve the resource at all.
  */
 public class ThemeIncludeEventHandler implements IncludeEventHandler {
@@ -60,40 +54,18 @@ public class ThemeIncludeEventHandler implements IncludeEventHandler {
 
         String path = includeResourcePath.trim();
 
-        if (isOutsideNamespace(path) || isNotATemplateName(path)) {
+        if (isOutsideNamespace(path)) {
             // Logged rather than raised: a template that asks for something it
             // may not have renders without that fragment, which is how Velocity
             // already treats a resource it cannot find.
-            LOG.warn("Refusing #" + directiveName + " of '" + path
+            LOG.debug("Refusing #" + directiveName + " of '" + path
                     + "' from '" + currentResourcePath + "': outside the template namespace");
             return null;
         }
 
-        return path;
-    }
-
-    /**
-     * Stored templates are resolved by id and carry no extension; theme
-     * resources are Velocity templates. A name bearing any other extension is
-     * asking for something that is not a template.
-     *
-     * @return true when the name is not one of those two shapes
-     */
-    private boolean isNotATemplateName(String path) {
-        // Stored template ids arrive as <template>|<deviceType>; the device
-        // type is a rendition selector, not part of the resource name.
-        String name = path;
-        int bar = name.indexOf('|');
-        if (bar > -1) {
-            name = name.substring(0, bar);
-        }
-
-        int dot = name.lastIndexOf('.');
-        if (dot == -1) {
-            // No extension: a stored template id.
-            return false;
-        }
-        return !name.regionMatches(true, dot, ".vm", 0, 3) || dot != name.length() - 3;
+        // Keep the caller's spelling (including any intentional surrounding
+        // whitespace) once validation has accepted the identifier.
+        return includeResourcePath;
     }
 
     /**
@@ -110,11 +82,21 @@ public class ThemeIncludeEventHandler implements IncludeEventHandler {
         if (normalized.contains("../") || normalized.endsWith("..")) {
             return true;
         }
-        // A colon before any slash indicates a scheme or a Windows drive.
+        // URI schemes and Windows drive prefixes are not theme identifiers.
         int colon = normalized.indexOf(':');
         if (colon > -1) {
-            int slash = normalized.indexOf('/');
-            return slash == -1 || colon < slash;
+            if (normalized.indexOf(':', colon + 1) > -1) {
+                return true;
+            }
+            if (colon == 1 && Character.isLetter(normalized.charAt(0))) {
+                return true;
+            }
+            if (colon + 1 == normalized.length()
+                    || normalized.charAt(colon + 1) == '/') {
+                return true;
+            }
+            String namespace = normalized.substring(0, colon);
+            return namespace.indexOf('.') > -1 || namespace.indexOf('/') > -1;
         }
         return false;
     }
--- a/app/src/main/java/org/apache/roller/weblogger/ui/rendering/velocity/ThemeIncludeEventHandler.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/rendering/velocity/ThemeIncludeEventHandler.java
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  The ASF licenses this file to You
+ * under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/apache/roller/weblogger/ui/rendering/velocity/ThemeIncludeEventHandler.java
```diff
@@ -0,0 +1,121 @@
...
+            return true;
+        }
+        if (normalized.contains("../") || normalized.endsWith("..")) {
+            return true;
+        }
+        // A colon before any slash indicates a scheme or a Windows drive.
+        int colon = normalized.indexOf(':');
+        if (colon > -1) {
+            int slash = normalized.indexOf('/');
+            return slash == -1 || colon < slash;
```
mraible: This refuses every shared-theme template id. `SharedThemeFromDir` builds ids as `themeId + ":" + templateName` (lines 270, 368) and `weblog.vm`'s `includeTemplate` macro does `#parse($pageArg.id + '|' + $model.deviceType)`, so the basic theme's weblog page asks for `basic:_day|standard`, gets `null` back, and renders nothing for that fragment. Every bundled theme is affected; only custom (UUID-id) templates pass. A scheme is followed by `/` (`file:///`, `http://`, `jar:file:`) or is a single drive letter, so something like `colon > 1 && (slash == -1 || colon < slash) && normalized.charAt(colon + 1) != '/'` is closer, and `ThemeResourceLoader` is the real authority on what a theme id looks like.

app/src/main/java/org/apache/roller/weblogger/ui/rendering/velocity/ThemeIncludeEventHandler.java
```diff
@@ -0,0 +1,121 @@
...
+        if (bar > -1) {
+            name = name.substring(0, bar);
+        }
+
+        int dot = name.lastIndexOf('.');
+        if (dot == -1) {
+            // No extension: a stored template id.
+            return false;
+        }
+        return !name.regionMatches(true, dot, ".vm", 0, 3) || dot != name.length() - 3;
```
mraible: Shared-theme template names are free-form: the basic theme's stylesheet template is literally named `basic-custom.css`, and the template guide documents inlining it through `#includeTemplate`. This check strips `|standard`, sees `.css`, and refuses it. Since the resource loaders are already confined to the theme and webapp namespaces, I'd drop the extension check rather than try to enumerate template-name shapes.

app/src/test/java/org/apache/roller/weblogger/ui/rendering/velocity/ThemeIncludeConfinementTest.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /**
+     * The shapes Roller itself includes must still pass: a stored template
+     * resolved by id, and the feed templates the servlets name directly.
+     */
+    @Test
+    public void legitimateIncludesStillPass() {
+        ThemeIncludeEventHandler handler = new ThemeIncludeEventHandler();
+        String[] allowed = {
```
mraible, line 139: None of these are the shape Roller's own macros pass to `#parse`. Add `basic:_day|standard` and `basic:basic-custom.css|standard` here; both fail against the current handler and would have caught the regression.

app/src/main/java/org/apache/roller/weblogger/ui/rendering/velocity/ThemeIncludeEventHandler.java
```diff
@@ -0,0 +1,121 @@
...
+            return null;
+        }
+
+        String path = includeResourcePath.trim();
+
+        if (isOutsideNamespace(path) || isNotATemplateName(path)) {
+            // Logged rather than raised: a template that asks for something it
+            // may not have renders without that fragment, which is how Velocity
+            // already treats a resource it cannot find.
+            LOG.warn("Refusing #" + directiveName + " of '" + path
```
mraible: Once the id bug is fixed this is fine, but note it's unthrottled: any weblog admin can make the server log a WARN per request by leaving a refused include in a public template. Debug, or a once-per-template warning, would be safer.

app/src/main/webapp/WEB-INF/velocity.properties
```diff
@@ -15,7 +15,11 @@
 # directory of this distribution.
 
 # specify resource loaders to use
-resource.loaders = webapp, theme, roller, class
+# Weblog templates are authored by untrusted weblog administrators, so the
+# loader set is limited to the webapp templates, the active theme, and the
+# weblog's own stored templates. The classpath is deliberately not a
+# resolvable namespace for them.
+resource.loaders = webapp, theme, roller
```
mraible, line 22: With the `class` loader gone, `RollerVelocity` (lines 66-67) still sets `resource.loader.class.cache` and `resource.loader.class.modification_check_interval` under `themes.reload.mode`; those two lines should go in the same change.

app/src/test/resources/WEB-INF/velocity.properties
```diff
@@ -70,3 +68,9 @@ velocimacro.inline.local_scope=false
 # set encoding/charset to UTF-8
 resource.default_encoding=UTF-8
 default.contentType=text/html; charset=utf-8
+
+# Weblog templates render under SecureUberspector, which governs method access
+# rather than resource resolution, so the include directives are confined
+# separately. Keep this aligned with /WEB-INF/velocity.properties.
```
mraible: The commit message says nothing reads this copy. Rather than keep two hand-maintained copies of security-relevant config plus a test that scans both, I'd delete this one.

app/src/test/java/org/apache/roller/weblogger/ui/rendering/velocity/ThemeIncludeConfinementTest.java
```diff
@@ -0,0 +1,215 @@
...
+     * still admits the classpath is a copy that can quietly become live.
+     */
+    private static final Path[] VELOCITY_PROPERTIES = {
+            Paths.get("src", "main", "webapp", "WEB-INF", "velocity.properties"),
+            Paths.get("src", "test", "resources", "WEB-INF", "velocity.properties"),
+    };
+
+    private String read(Path path) throws Exception {
+        assertTrue(Files.isReadable(path),
+                "cannot read " + path.toAbsolutePath() + " (run from the app module)");
```
mraible, line 61: cwd-relative, so this only runs from `app/`; surefire sets `project.build.directory` for this module (see `ApplicationResourcesTest`), which would let it run from an IDE rooted at the repo.

app/src/test/java/org/apache/roller/weblogger/ui/rendering/velocity/ThemeIncludeConfinementTest.java
```diff
@@ -0,0 +1,215 @@
...
+     * <p>Velocity's ClasspathResourceLoader resolves a plain resource name
+     * against the classpath, with no traversal involved, so a loader set that
+     * includes it makes any packaged file resolvable by name. The first case
+     * reproduces that resolution, which is what gives the other two something
+     * to be measured against: each of the two changes is then shown to stop it
+     * on its own, so neither is carrying the other.
+     */
+    @Test
+    public void aPlainNameDoesNotReachAPackagedFile() throws Exception {
+        Path dir = Files.createTempDirectory("roller-include-confinement");
```
mraible, line 166: Nit: the temp directory is never deleted; a `@TempDir` parameter does the cleanup.

app/src/main/java/org/apache/roller/weblogger/ui/rendering/velocity/ThemeIncludeEventHandler.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Override
+    public String includeEvent(Context context, String includeResourcePath,
+                               String currentResourcePath, String directiveName) {
+
+        if (includeResourcePath == null || includeResourcePath.trim().isEmpty()) {
+            return null;
+        }
+
+        String path = includeResourcePath.trim();
```
mraible, line 55: Nit: `trim()` is computed twice, and returning the trimmed value means `#parse(" $pageId")` now resolves a different name than before. Compute it once and return the original unless trimming is intentional.

mraible (commented):
Reviewed. The resource-loader and namespace changes are moving in the right direction. Please address shared-theme compatibility before merge along with the remaining cleanup and test notes.

app/src/test/resources/WEB-INF/velocity.properties
```diff
@@ -70,3 +68,9 @@ velocimacro.inline.local_scope=false
 # set encoding/charset to UTF-8
 resource.default_encoding=UTF-8
 default.contentType=text/html; charset=utf-8
+
+# Weblog templates render under SecureUberspector, which governs method access
```
mbien, line 72: that name :)

mbien (approved):

PR #173, opened 2026-08-29
Use a shared JDOM builder for bookmark and configuration parsing

snoopdave:
Several call sites each construct their own JDOM SAXBuilder with inconsistent parser settings. This change consolidates them behind one shared, consistently configured builder.

## What changed

- Add SafeSAXBuilder with secure processing enabled, document type declarations disallowed, external entity and DTD resolution disabled, and entity expansion disabled.
- Apply it to OPML bookmark import and the retained internal parsers: MenuHelper, RuntimeConfigDefsParser, and ThemeMetadataParser.
- Report unsupported declarations during bookmark import with a concise input error.
- Leave Trackback.java untouched because PR #163 removes that file.

## Merge order

Merge after #163 so no older parser call site remains.

## Release note

OPML and hand-written theme.xml documents carrying a DOCTYPE are now rejected; shipped themes do not use one.

## Tests

SafeSAXBuilderTest asserts the shared parser contract. BookmarkImportParsingTest loads its fixture from the classpath, verifies ordinary OPML import, and checks the typed error and retained parse root cause for rejected declarations.

Local Temurin JDK 11 verification: 4 focused tests passed; the full reactor passed 162 tests with 1 skipped.

### DIFF
--- a/app/src/main/java/org/apache/roller/weblogger/business/jpa/JPABookmarkManagerImpl.java
+++ b/app/src/main/java/org/apache/roller/weblogger/business/jpa/JPABookmarkManagerImpl.java
@@ -34,6 +34,7 @@ import org.apache.roller.weblogger.pojos.WeblogBookmarkFolder;
 import org.apache.roller.weblogger.pojos.Weblog;
 import org.jdom2.Document;
 import org.jdom2.Element;
+import org.jdom2.input.JDOMParseException;
 import org.apache.roller.weblogger.util.SafeSAXBuilder;
 
 /*
@@ -157,6 +158,10 @@ public class JPABookmarkManagerImpl implements BookmarkManager {
             for (Element elem : body.getChildren()) {
                 importOpmlElement(elem, newFolder );
             }
+        } catch (JDOMParseException ex) {
+            throw new WebloggerException(
+                    "Unable to import bookmarks: XML document type declarations are not supported",
+                    ex);
         } catch (Exception ex) {
             throw new WebloggerException(ex);
         }
--- a/app/src/main/java/org/apache/roller/weblogger/business/jpa/JPABookmarkManagerImpl.java
+++ b/app/src/main/java/org/apache/roller/weblogger/business/jpa/JPABookmarkManagerImpl.java
@@ -34,7 +34,7 @@ import org.apache.roller.weblogger.pojos.WeblogBookmarkFolder;
 import org.apache.roller.weblogger.pojos.Weblog;
 import org.jdom2.Document;
 import org.jdom2.Element;
-import org.jdom2.input.SAXBuilder;
+import org.apache.roller.weblogger.util.SafeSAXBuilder;
 
 /*
  * JPABookmarkManagerImpl.java
@@ -142,7 +142,7 @@ public class JPABookmarkManagerImpl implements BookmarkManager {
 
         try {
             // Build JDOC document OPML string
-            SAXBuilder builder = new SAXBuilder();
+            SafeSAXBuilder builder = new SafeSAXBuilder();
             StringReader reader = new StringReader( opml );
             Document doc = builder.build( reader );
 
--- a/app/src/main/java/org/apache/roller/weblogger/business/themes/ThemeMetadataParser.java
+++ b/app/src/main/java/org/apache/roller/weblogger/business/themes/ThemeMetadataParser.java
@@ -25,7 +25,7 @@ import org.apache.roller.weblogger.pojos.ThemeTemplate.ComponentType;
 import org.jdom2.Document;
 import org.jdom2.Element;
 import org.jdom2.JDOMException;
-import org.jdom2.input.SAXBuilder;
+import org.apache.roller.weblogger.util.SafeSAXBuilder;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -52,7 +52,7 @@ public class ThemeMetadataParser {
         
         ThemeMetadata theme = new ThemeMetadata();
         
-        SAXBuilder builder = new SAXBuilder();
+        SafeSAXBuilder builder = new SafeSAXBuilder();
         Document doc = builder.build(instream);
         
         // start at root and get theme id, name, description and author
--- a/app/src/main/java/org/apache/roller/weblogger/util/SafeSAXBuilder.java
+++ b/app/src/main/java/org/apache/roller/weblogger/util/SafeSAXBuilder.java
@@ -1,0 +1,127 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  The ASF licenses this file to You
+ * under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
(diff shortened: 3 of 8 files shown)


app/src/main/java/org/apache/roller/weblogger/util/SafeSAXBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * <p>An XML document can name resources for the parser to go and read: a
+ * document type declaration can point at an external subset, and entity
+ * declarations can point at files or URLs. Resolving those makes the parser act
+ * on behalf of whoever wrote the document, which is only appropriate when the
+ * document is Roller's own.
+ *
+ * <p>Roller parses documents from user input and from its own menu, theme and
+ * configuration descriptors alike. Rather than track which parser is on which
+ * side, every retained JDOM parser is built here, and none of them resolve
```
mraible: Not quite true yet: `Trackback.parseTrackbackResponse` (`Trackback.java:177`) still uses a bare `new SAXBuilder()` on a remote server's response. #163 deletes that file, so either merge this after #163 or switch that call site here too.

app/src/main/java/org/apache/roller/weblogger/business/themes/ThemeMetadataParser.java
```diff
@@ -52,7 +52,7 @@ public ThemeMetadata unmarshall(InputStream instream)
         
         ThemeMetadata theme = new ThemeMetadata();
         
-        SAXBuilder builder = new SAXBuilder();
+        SafeSAXBuilder builder = new SafeSAXBuilder();
```
mraible, line 55: A `theme.xml` with a DOCTYPE is now refused, and `ThemeManagerImpl.loadAllThemesFromDisk` just logs "Problem processing theme", so the theme vanishes and its weblogs throw `ThemeNotFoundException`. Shipped themes have no DOCTYPE, so this only hits hand-written ones, but please mention it with the OPML note.

app/src/main/java/org/apache/roller/weblogger/business/jpa/JPABookmarkManagerImpl.java
```diff
@@ -142,7 +142,7 @@ public void importBookmarks(
 
         try {
             // Build JDOC document OPML string
-            SAXBuilder builder = new SAXBuilder();
+            SafeSAXBuilder builder = new SafeSAXBuilder();
```
mraible, line 146: `BookmarksImport` shows `ex.toString()`, so a user whose OPML export has `<!DOCTYPE opml>` now sees `org.apache.roller.weblogger.WebloggerException: org.jdom2.input.JDOMParseException: ... DOCTYPE is disallowed when the feature ...`. Catching `JDOMParseException` here and wrapping it with a message like "OPML files with a DOCTYPE are not accepted" would tell them what to do.

app/src/test/java/org/apache/roller/weblogger/business/BookmarkImportParsingTest.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+        return folder.retrieveBookmarks();
+    }
+
+    private void tryImport(String opml) {
+        try {
+            bookmarkManager().importBookmarks(
+                    TestUtils.getManagedWebsite(testWeblog), folderName, opml);
+            TestUtils.endSession(true);
+        } catch (Exception expected) {
```
mraible, line 88: This catches everything, so the DOCTYPE test passes whenever the import fails for any reason (DB state, a `getFolder` regression), and the session is left un-ended when `importBookmarks` throws. Assert on `WebloggerException` with a `JDOMParseException` cause, and end the session in a `finally`.

app/src/test/java/org/apache/roller/weblogger/business/BookmarkImportParsingTest.java
```diff
@@ -0,0 +1,122 @@
...
+            // each test say what must be true either way.
+            log.debug("import raised: " + expected);
+        }
+    }
+
+    /** Ordinary OPML, with no declarations in it, must still import. */
+    @Test
+    public void ordinaryOpmlStillImports() throws Exception {
+        byte[] opml = Files.readAllBytes(
+                new File("src/test/resources/bookmarks.opml").toPath());
```
mraible: cwd-relative; `BookmarkTest` and `FileContentManagerTest` load the same fixture from the classpath.

app/src/main/java/org/apache/roller/weblogger/util/SafeSAXBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+     * where a parser that does not recognise them can be tolerated.
+     *
+     * <p>They are JAXP properties rather than SAX ones, and Roller ships its
+     * own Xerces, which rejects them outright at the SAX layer. Setting them
+     * through the builder would therefore fail every parse. They are still
+     * worth setting where they are understood, because they deny the protocols
+     * outright, so they are applied here and a rejection is logged and passed
+     * over — the features above are what carry the guarantee.
+     */
+    private static final class HardenedReaders implements XMLReaderJDOMFactory {
```
mraible: On this classpath `SAXParserFactory.newInstance()` resolves to Xerces 2.11 (via `nekohtml`), which throws `SAXNotRecognizedException` for both `ACCESS_EXTERNAL_DTD` and `ACCESS_EXTERNAL_SCHEMA`, so `denyProtocol()` always takes the swallowed-exception branch. With DOCTYPE refused there's nothing for those properties to restrict anyway; `XMLReaders.NONVALIDATING` plus the feature calls above is enough and this inner class can go.

app/src/main/java/org/apache/roller/weblogger/util/SafeSAXBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+
+        // A document that declares a doctype is refused. Everything an entity
+        // could name has to be declared first, so this is the setting the rest
+        // stand behind.
+        setFeature(DISALLOW_DOCTYPE, true);
+
+        setFeature(EXTERNAL_GENERAL_ENTITIES, false);
+        setFeature(EXTERNAL_PARAMETER_ENTITIES, false);
+        setFeature(LOAD_EXTERNAL_DTD, false);
+
```
mraible: Nit: in jdom2 `setExpandEntities(false)` writes the same feature key as `setFeature(EXTERNAL_GENERAL_ENTITIES, false)` four lines up; one of the two is a no-op.

mraible (approved):
Reviewed. Please ensure this merges after #163 or update the remaining call site, document the theme compatibility change, and refresh the test description.

app/src/main/java/org/apache/roller/weblogger/util/SafeSAXBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+ * side, every retained JDOM parser is built here, and none of them resolve
+ * anything. Roller's own descriptors carry no document type declaration, so the
+ * strict setting costs them nothing.
+ *
+ * <p>The settings overlap deliberately. Refusing the declaration outright is
+ * what does the work; the remaining ones close the same door at the layers
+ * beneath, so a parser configured elsewhere, or a JAXP implementation with
+ * different defaults, does not quietly reopen it.
+ */
+public class SafeSAXBuilder extends SAXBuilder {
```
mbien, line 46: interesting that Roller doesn't have `DocumentBuilderFactory` anywhere. It was brand new in 2002 :)

mbien (commented):
this could also set FEATURE_SECURE_PROCESSING most likely.

https://github.com/apache/roller/blob/f1422f4448f9ff502600069a1c316bafea02b3d7/app/src/main/java/org/apache/roller/weblogger/business/WebloggerImpl.java#L367-L378

tests are green with it

PR #174, opened 2026-08-29
Derive media content types from file content, not the request

snoopdave:
This change derives a stored media file's content type from the file itself
rather than from the type declared with the upload, and serves media inline only
for a small explicit allow-list of passive formats.

## What changed

- Add `MediaTypePolicy` as the single place for the stored type, the inline
  allow-list, and the response headers.
- Derive the stored type from the filename; consult the declared type only for
  opaque names and never adopt an active type (an explicit list plus any `+xml`
  suffix).
- Send `X-Content-Type-Options: nosniff` on every media response.
- Serve inline only passive images, audio, video, and PDF; serve everything else
  as `application/octet-stream` with an attachment disposition. SVG is excluded.
- Route all five upload entry points and the three serving paths through the
  policy.

Note for the release notes: media held as CSS or JavaScript now downloads
instead of loading inline — a user-visible compatibility change.

## Tests

`MediaTypePolicyTest` (9 behavioral cases plus 3 source audits asserting every
caller routes through the policy) covers a file whose declared type does not
match its name, SVG / XHTML and unknown types, genuine images retaining their
type, and `nosniff` on every response.

### DIFF
--- a/app/src/main/java/org/apache/roller/weblogger/ui/rendering/servlets/PreviewResourceServlet.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/rendering/servlets/PreviewResourceServlet.java
@@ -163,8 +163,8 @@ public class PreviewResourceServlet extends HttpServlet {
         }
 
         // set the content type based on whatever is in our web.xml mime defs
-        String resourceType = this.context.getMimeType(
-                resourceRequest.getResourcePath());
+        String resourceType = MediaTypePolicy.typeFromName(
+                resourceRequest.getResourcePath(), this.context::getMimeType);
         if (fromUploadedMedia) {
             // Uploaded through the media library, so it is governed by the
             // same policy as any other media response.
--- a/app/src/main/java/org/apache/roller/weblogger/ui/rendering/servlets/PreviewResourceServlet.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/rendering/servlets/PreviewResourceServlet.java
@@ -29,6 +29,7 @@ import javax.servlet.http.HttpServletResponse;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.roller.weblogger.util.MediaTypePolicy;
 import org.apache.roller.weblogger.WebloggerException;
 import org.apache.roller.weblogger.business.MediaFileManager;
 import org.apache.roller.weblogger.business.WebloggerFactory;
@@ -129,7 +130,9 @@ public class PreviewResourceServlet extends HttpServlet {
         }
 
         // if not from theme then see if resource is in weblog's upload dir
+        boolean fromUploadedMedia = false;
         if (resourceStream == null) {
+            fromUploadedMedia = true;
             try {
                 MediaFileManager mmgr = WebloggerFactory.getWeblogger()
                         .getMediaFileManager();
@@ -160,8 +163,19 @@ public class PreviewResourceServlet extends HttpServlet {
         }
 
         // set the content type based on whatever is in our web.xml mime defs
-        response.setContentType(this.context.getMimeType(resourceRequest
-                .getResourcePath()));
+        String resourceType = this.context.getMimeType(
+                resourceRequest.getResourcePath());
+        if (fromUploadedMedia) {
+            // Uploaded through the media library, so it is governed by the
+            // same policy as any other media response.
+            MediaTypePolicy.applyResponseHeaders(response, resourceType,
+                    resourceRequest.getResourcePath());
+        } else {
+            // A theme resource: authored as part of the theme and served as
+            // the type the theme intends, but never re-typed by the browser.
+            response.setHeader("X-Content-Type-Options", "nosniff");
+            response.setContentType(resourceType);
+        }
 
         try {
             // ok, lets serve up the file
--- a/app/src/main/java/org/apache/roller/weblogger/ui/rendering/servlets/ResourceServlet.java
+++ b/app/src/main/java/org/apache/roller/weblogger/ui/rendering/servlets/ResourceServlet.java
@@ -30,6 +30,7 @@ import javax.servlet.http.HttpServletResponse;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.roller.weblogger.util.MediaTypePolicy;
 import org.apache.roller.weblogger.WebloggerException;
 import org.apache.roller.weblogger.business.MediaFileManager;
 import org.apache.roller.weblogger.business.WebloggerFactory;
@@ -125,7 +126,9 @@ public class ResourceServlet extends HttpServlet {
         }
 
         // if not from theme then see if resource is in weblog's upload dir
+        boolean fromUploadedMedia = false;
         if (resourceStream == null) {
+            fromUploadedMedia = true;
             try {
                 MediaFileManager mmgr = WebloggerFactory.getWeblogger()
                         .getMediaFileManager();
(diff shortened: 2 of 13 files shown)


app/src/main/java/org/apache/roller/weblogger/ui/rendering/servlets/ResourceServlet.java
```diff
@@ -159,8 +162,19 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
         }
 
         // set the content type based on whatever is in our web.xml mime defs
-        response.setContentType(this.context.getMimeType(resourceRequest
-                .getResourcePath()));
+        String resourceType = this.context.getMimeType(
+                resourceRequest.getResourcePath());
+        if (fromUploadedMedia) {
+            // Uploaded through the media library, so it is governed by the
```
mraible, line 168: This branch is also where customized-theme resources land: `WeblogCustomTheme.getResource()` looks up the media file but never assigns it to `resource`, so it always returns `null`, and `importTheme` stores theme CSS/JS as media files with their original path. `resourceType` for `css/bootstrap.css` is `text/css`, which isn't inline-safe, so the stylesheet goes out as an octet-stream attachment with `nosniff` and the browser drops it. Fixing `WeblogCustomTheme.getResource` to return the media file would route these through the theme branch above; alternatively, treat `text/css` / `text/javascript` from the servlet-context mime table as theme-authored here as the description already promises.

app/src/main/java/org/apache/roller/weblogger/util/MediaTypePolicy.java
```diff
@@ -0,0 +1,186 @@
...
+    private MediaTypePolicy() {
+    }
+
+    public static final String DEFAULT_TYPE = "application/octet-stream";
+
+    /**
+     * Formats browsers render without executing anything the file carries.
+     * SVG is deliberately absent: it is an XML document that can carry script.
+     */
+    private static final Set<String> INLINE_TYPES = Collections.unmodifiableSet(
```
mraible: Worth listing what this drops compared to `response.setContentType(mediaFile.getContentType())` on master. Existing `image/svg+xml` media (already stored, already embedded in entries with ``) now downloads instead of rendering, since browsers never sniff SVG; and `image/x-png` (legacy IE uploads), `image/jpg` (some XML-RPC clients), `image/avif` and `image/apng` are all passive but absent here. If SVG stays out (defensible, it can carry script), the release note should say so explicitly, and I'd add the other image subtypes.

app/src/main/java/org/apache/roller/weblogger/util/MediaTypePolicy.java
```diff
@@ -0,0 +1,186 @@
...
+        String declared = normalize(declaredType);
+        if (isConclusive(declared) && !isActive(declared)) {
+            return declared;
+        }
+
+        return DEFAULT_TYPE;
+    }
+
+    /** @return true when browsers render this type without executing it */
+    public static boolean isInlineSafe(String contentType) {
```
mraible: `text/plain` with `nosniff` can't execute anything, so uploaded `.txt` files (very common as attachments) could stay inline rather than becoming a save-as dialog for an `application/octet-stream` `notes.txt`.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/MediaFileEdit.java
```diff
@@ -124,7 +125,10 @@ public String save() {
 
                 if (uploadedFile != null) {
                     mediaFile.setLength(this.uploadedFile.length());
-                    mediaFile.setContentType(this.uploadedFileContentType);
+                    // Replacing the body re-decides the type, on the same
+                    // terms as the original upload.
+                    mediaFile.setContentType(MediaTypePolicy.storedTypeFor(
+                            mediaFile.getName(), this.uploadedFileContentType));
```
mraible: On replace this derives the type from the record's (old) name rather than the uploaded replacement's name (`this.uploadedFileName` is right there). Replacing `photo.jpg` with `photo.png` without renaming stores `image/jpeg` for PNG bytes and serves them with `nosniff`; replacing `clip.txt` with a video keeps `text/plain` and forces a download. `putMedia` in `MediaCollection` has the same mismatch with `mf.getName()`.

app/src/main/java/org/apache/roller/weblogger/util/MediaTypePolicy.java
```diff
@@ -0,0 +1,186 @@
...
+                    "text/xsl", "application/xslt+xml")));
+
+    /**
+     * @param fileName      the uploaded file's name
+     * @param declaredType  the type the client said it was, may be null
+     * @return the type to store: derived from the name where that is
+     *         conclusive, otherwise the declared type if it is not one a
+     *         browser would act on, otherwise the generic binary type
+     */
+    public static String storedTypeFor(String fileName, String declaredType) {
```
mraible: Because active declared types collapse to `application/octet-stream` *before* `canSave` / `checkFileType` run, an admin's content-type rules in `uploads.types.forbid` (e.g. `image/svg+xml,application/xhtml+xml`) no longer match those uploads; the file is stored under the generic type instead of being refused. Run the forbid check against the declared type too, or document that forbid rules are now extension-based.

app/src/main/java/org/apache/roller/weblogger/util/MediaTypePolicy.java
```diff
@@ -0,0 +1,186 @@
...
+                "attachment; filename=\"" + headerSafe(fileName) + "\"");
+    }
+
+    private static String deriveFromName(String fileName) {
+        if (fileName == null || fileName.trim().isEmpty()) {
+            return null;
+        }
+        try {
+            return Utilities.getContentTypeFromFileName(fileName);
+        } catch (Exception undetermined) {
```
mraible: `Utilities.getContentTypeFromFileName` is backed by `javax.activation`'s default map, which knows about 22 extensions (no `pdf`, `svg`, `webp`, `mp4`, `mp3`, `zip`, `css`, `js`). So "the name decides" only holds for those; for everything else the declared type is stored after all (a `report.pdf` declared `application/zip` is stored as zip and downloads). The servlets already use the servlet context's mime table (`web.xml` has a full one); the policy should consult the same table.

app/src/main/java/org/apache/roller/weblogger/webservices/xmlrpc/MetaWeblogAPIHandler.java
```diff
@@ -381,7 +382,7 @@ public Object newMediaObject(String blogid, String userid, String password,
             mf.setDirectory(root);
             mf.setWeblog(website);
             mf.setName(name);
-            mf.setContentType(type);
+            mf.setContentType(MediaTypePolicy.storedTypeFor(name, type));
```
mraible: Previously `mf.setContentType(null)` made `checkFileType` reject an upload with no `type` member; `storedTypeFor(name, null)` now always returns a type, so the upload is accepted as `application/octet-stream` when the allow list is empty (the default). Probably fine, but it's a loosening the description should mention.

app/src/main/java/org/apache/roller/weblogger/ui/rendering/servlets/PreviewResourceServlet.java
```diff
@@ -160,8 +163,19 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
 
         // set the content type based on whatever is in our web.xml mime defs
-        response.setContentType(this.context.getMimeType(resourceRequest
-                .getResourcePath()));
+        String resourceType = this.context.getMimeType(
+                resourceRequest.getResourcePath());
+        if (fromUploadedMedia) {
+            // Uploaded through the media library, so it is governed by the
+            // same policy as any other media response.
+            MediaTypePolicy.applyResponseHeaders(response, resourceType,
```
mraible, line 171: `context.getMimeType()` returns `null` for extensions not mapped in `web.xml`/the container, and `applyResponseHeaders` turns `null` into an octet-stream attachment. On master `setContentType(null)` left the type unset and the browser could still display the file; now any unmapped (or uppercase, on a case-sensitive container) extension downloads in the theme preview.

mraible (commented):
Reviewed. Centralizing media-type handling is a good direction. Please address the customized-theme regression and the remaining protocol and compatibility notes before merge.

app/src/main/java/org/apache/roller/weblogger/ui/rendering/servlets/PreviewResourceServlet.java
mbien: > or uppercase, on a case-sensitive container

the name is normalized before the method reference of `getMimeType(name)` is called. So at least case should not matter, no?

mbien: sorry, I just realized a commit was added _after_ this comment was posted.

mbien (approved):

PR #175, opened 2026-08-30
Store enclosure metadata without fetching remote URLs

snoopdave:
Summary:
- collect enclosure media type and byte length alongside its URL
- validate HTTP(S) enclosure metadata locally before saving
- populate enclosure metadata from uploaded media files
- remove the legacy remote metadata lookup classes

Testing:
- mvn -pl app -Dtest=EnclosureMetadataTest test
- mvn -V -ntp install (Temurin JDK 11.0.31)


app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/EntryEdit.java
```diff
@@ -192,6 +190,19 @@ public String publish() {
...
         if (!hasActionErrors()) {
+            EnclosureMetadata enclosure = null;
+            if (!StringUtils.isEmpty(getBean().getEnclosureURL())) {
+                try {
+                    enclosure = EnclosureMetadata.of(
+                            getBean().getEnclosureURL(),
+                            getBean().getEnclosureType(),
+                            getBean().getEnclosureLength());
+                } catch (IllegalArgumentException e) {
+                    addError("weblogEdit.enclosureMetadataInvalid");
```
mraible: This is fatal where the old flow was advisory (`addMessage` and continue), and existing entries can't pass it: the removed `MediacastUtil` stored `con.getContentType()` verbatim, so `att_mediacast_type` may be `audio/mpeg; charset=utf-8` or `video/mp4;codecs=avc1`, which `MEDIA_TYPE` rejects. The author then can't save any change to that entry until they notice and hand-edit the type. Either accept parameters in the regex (and strip them), or treat an invalid legacy value as "clear the enclosure and warn" instead of refusing the save. Also, one generic message for three fields: a blank Length (new field, previously auto-filled) produces the same text as a bad URL.

app/src/main/java/org/apache/roller/weblogger/util/EnclosureMetadata.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.util.regex.Pattern;
+import org.apache.commons.validator.routines.UrlValidator;
+
+/**
+ * Validated metadata for an RSS or Atom enclosure.
+ */
+public final class EnclosureMetadata {
+
+    private static final UrlValidator URL_VALIDATOR = new UrlValidator(
```
mraible: `UrlValidator` rejects any host whose TLD isn't in its IANA list (`blog.internal`, `roller.local`, `blog.lan`, `.test`), underscores, and IDN hosts; `ALLOW_LOCAL_URLS` only admits single-label hosts like `localhost`. Since `MediaFileAddSuccess` pre-fills the enclosure URL from the site's own absolute URL, an intranet install can never attach uploaded media as an enclosure, and existing entries with such URLs can no longer be re-saved. `java.net.URI` with a scheme check (`http` / `https`, non-empty host) is enough here; the point is no longer fetching it.

app/src/main/java/org/apache/roller/weblogger/util/EnclosureMetadata.java
```diff
@@ -0,0 +1,84 @@
...
+            new String[] {"http", "https"}, UrlValidator.ALLOW_LOCAL_URLS);
+
+    private static final Pattern MEDIA_TYPE = Pattern.compile(
```
mraible: This admits `&` (and `'`, `` ` ``, `|`), but ``feeds.vm`` lines 52 and 80 emit `type="$`mc_type`"` without ``escapeXML``, so `audio/mp&eg` passes validation and turns the whole RSS and Atom feed into malformed XML. Either escape it in ``feeds.vm`` or drop those characters from the character class (real media types never use them).

app/src/main/webapp/WEB-INF/jsps/editor/MediaFileAddSuccess.jsp
```diff
@@ -96,7 +98,9 @@
 
                         <div class="col-md-1">
                             <input type="radio" name="enclosure"
-                                   onchange="setEnclosure('<s:property value="%{#newFile.permalink}"/>')"/>
+                                   onchange="setEnclosure('<s:property value="%{#newFile.permalink}"/>',
+                                           '<s:property value="%{#newFile.contentType}"/>',
```
mraible: `#newFile.contentType` comes from the upload's declared `Content-Type` (`MediaFileAdd` only overrides `null` / `octet-stream`), and `<s:property>` HTML-escapes without touching `'`, so a part declared as `audio/x'; alert(1)//` runs in the uploader's session, and a stray quote just kills the handler so the radio does nothing. Use `escapeJavaScript="true" escapeHtml="false"`, or better, put the three values in `data-` attributes and read them in the handler. (#174 changes how `contentType` is derived, which narrows this, but the JSP should still not inline it unescaped.)

app/src/main/java/org/apache/roller/weblogger/ui/struts2/editor/EntryEdit.java
```diff
@@ -192,6 +190,19 @@ public String publish() {
...
+                    return INPUT;
```
mraible: This returns before the `if ("entryAdd".equals(actionName)) getBean().setStatus(null)` reset at the end of the method (line 309), which every other failed save on a new entry goes through. `publish()` has already stamped `PUBLISHED` on the bean, so the form re-renders with the green "Published (Last updated: )" badge and an empty date for an entry that was never written, and the hidden `bean.status` carries `PUBLISHED` into the next submit.

app/src/main/java/org/apache/roller/weblogger/util/EnclosureMetadata.java
```diff
@@ -0,0 +1,84 @@
...
+        String normalizedUrl = normalize(url);
+        String normalizedType = normalize(contentType);
+        String normalizedLength = normalize(length);
+
+        if (!URL_VALIDATOR.isValid(normalizedUrl)) {
+            throw new IllegalArgumentException("Enclosure URL must be an absolute HTTP or HTTPS URL");
+        }
+        if (!MEDIA_TYPE.matcher(normalizedType).matches()) {
+            throw new IllegalArgumentException("Enclosure type must be a valid media type");
+        }
```
mraible, line 55: Nit: no upper bound; `Long.MAX_VALUE` is accepted and the feed advertises an 8 EiB enclosure. The form caps the field at 20 characters, so a sanity ceiling here would match.

app/src/main/resources/ApplicationResources.properties
```diff
@@ -1606,13 +1606,12 @@ weblogEdit.trackbackError404=Trackback failed, could not reach trackback URL.  A
...
 
 weblogEdit.enclosureURL=Enclosure URL
-weblogEdit.enclosureURL.tooltip=Podcast or other multimedia URL to embed within the RSS & Atom feeds for this blog entry.
-weblogEdit.enclosureType=Type
-weblogEdit.enclosureLength=Length
-weblogEdit.mediaCastFailedFetchingInfo=Unable to reach the enclosure. Check the hostname in the URL.
-weblogEdit.mediaCastUrlMalformed=The enclosure URL was malformed.
-weblogEdit.mediaCastResponseError=The enclosure server returned an error. Do you have the right URL?
-weblogEdit.mediaCastLacksContentTypeOrLength=Unable to use enclosure URL. Server provided no content type or no length.
+weblogEdit.enclosureURL.tooltip=Absolute HTTP or HTTPS URL to embed within the RSS & Atom feeds for this blog entry.
```
mraible, line 1601: Nit: the four removed `weblogEdit.mediaCast*` keys are still in the `_de` / `_es` / `_fr` / `_ja` / `_ko` / `_ru` / `_zh_CN` bundles, and the `ja` / `zh_CN` tooltips still describe the old "podcast URL" semantics rather than the HTTP(S)-only requirement that now produces the error.

mraible (commented):
Reviewed. Please address the validation compatibility issues and output-handling notes before merge; several existing entries and intranet installations would otherwise be unable to save.

PR #176, opened 2026-08-30
Retire incoming Trackback support and normalize comment links

snoopdave:
Summary:
- retire incoming Trackback routing, runtime controls, and autodiscovery output
- retain deprecated no-op template, URL helper, and validator compatibility for one release
- centralize comment-author URL handling for posting and rendering, including local and internationalized hosts
- show invalid legacy author URLs to moderators as escaped text rather than links
- remove stale documentation while restoring localized comment-settings headings

Dependency:
- stacked on #163; the updated base branch was merged normally

Testing:
- focused Java 11 regression tests: 6 passed
- full mvn clean verify on Temurin JDK 11.0.31: 166 passed, 1 skipped


mraible (commented):
Retiring incoming Trackback is overdue and the removal itself is thorough. What I'd like fixed before approving is the compatibility for existing sites (inline), and a few notes outside the diff:

- `TrackbackLinkbackCommentValidator` is deleted, and it's a documented entry for `comment.validator.classnames` in the user guide. `CommentValidationManager` builds the list with `Reflection.newInstances`, which throws on the first missing class and leaves the list empty, so an install that lists it loses *every* comment validator (banned words, excess links, Akismet) silently. Either keep a no-op class for a release or make the manager skip unknown validators with a warning.
- `docs/roller-template-guide.adoc:846` still documents `$url.trackback(anchor)` and line 1861 still documents `#showTrackbackAutodiscovery($entry)`; `docs/roller-user-guide.adoc:1282` and `docs/examples/configs/tomcat/mod_jk.conf:10` also still reference trackbacks.

app/src/main/webapp/WEB-INF/velocity/weblog.vm
```diff
@@ -114,30 +113,6 @@ Show RSS, Atom and RSD auto-discovery links as HTML link elements.
...
-    dc:subject="$entry.category.name"
-    dc:description="$entry.title"
-    dc:creator="$entry.creator.userName"
-    dc:date="$entry.pubTime" />
-</rdf:RDF>
--->
-#end
-#end
-
 #**
```
mraible, line 123: Every release through 6.1.5 shipped `basic/_day.vm` and `basicmobile/_day.vm` with `#showTrackbackAutodiscovery($entry)`, and anyone who clicked "Customize theme" has that copied into their DB templates. `runtime.strict_mode.enable` is off, so Velocity renders the now-undefined macro call as literal text under every entry. Keep the macro as a documented no-op (`#macro( showTrackbackAutodiscovery $entry )#end`) for at least a release, and the same goes for `$url.trackback(...)` in `URLModel`, which custom templates reference outside an `#if`.

app/src/main/java/org/apache/roller/weblogger/pojos/wrapper/WeblogEntryCommentWrapper.java
```diff
@@ -93,7 +93,7 @@ public String getEmail() {
      * Value is always html escaped.
      */
     public String getUrl() {
-        return StringEscapeUtils.escapeHtml4(this.pojo.getUrl());
+        return StringEscapeUtils.escapeHtml4(this.pojo.getSafeUrl());
```
mraible: `CommentServlet` stores `""` for a blank URL, so this used to return `""` and the javadoc promises a non-`null` escaped value; `escapeHtml4(null)` is `null`, and with strict mode off a custom template that writes `$comment.url` without an `isEmpty` guard now prints the literal `$comment.url` into `href` for every comment without a URL. Return `""` when `getSafeUrl()` is `null`.

app/src/main/webapp/WEB-INF/jsps/editor/Comments.jsp
```diff
@@ -273,12 +273,12 @@
                                             </s:else>
                                         </div>
 
-                                        <s:if test="#comment.url != null && !#comment.url.equals('')">
+                                        <s:if test="#comment.safeUrl != null">
```
mraible: Hiding the URL entirely when it fails validation takes away the thing the moderator needs to judge the comment (`javascript:...`, a homoglyph domain, an intranet host). Render the raw value as escaped text when `safeUrl` is `null` and only make it a link when it isn't.

app/src/main/java/org/apache/roller/weblogger/util/CommentAuthorUrl.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private CommentAuthorUrl() {
+    }
+
+    public static String normalize(String value) {
+        if (value == null || value.isBlank()) {
+            return null;
+        }
+        String normalized = value.trim();
+        return VALIDATOR.isValid(normalized) ? normalized : null;
```
mraible: Default `UrlValidator` (no `ALLOW_LOCAL_URLS`, IANA TLD list, ASCII path regex) rejects `http://localhost:8080/blog`, `http://intranet-wiki/page`, `http://my_host.example.com/` and IDN paths, so legacy comments and imported ones lose their author link on public pages and vanish from the admin page, with no log line. `CommentServlet` accepts on post with the same validator today, so at minimum reuse one shared instance/policy for both (this class is the natural home; have the servlet call `normalize`), and consider `ALLOW_LOCAL_URLS` for intranet installs.

app/src/main/resources/ApplicationResources_de.properties
```diff
@@ -252,19 +250,15 @@ configForm.allowNewUsers=Erlaube das Anlegen neuer Benutzer?
 configForm.allowedExtensions=Zul\u00E4ssige Dateierweiterungen
 configForm.commentHtmlAllowed=HTML in Kommentaren erlauben?
 configForm.commentPlugins=An-/Abschalten von Plugins zur Kommentarformatierung
-configForm.commentSettings=Kommentar und Trackback Einstellungen
 configForm.editorPages=Bearbeitungsseiten
 configForm.emailComments=E-Mailbenachrichtung bei Kommentaren?
 configForm.enableComments=Kommentare in Weblogs erlauben?
```
mraible, line 256: `configForm.commentSettings` was removed from all seven translated bundles but is still the display-group key in `runtimeConfigDefs.xml:155`, so the Global Configuration heading falls back to English for every non-English admin. Re-add it (minus the Trackback wording) in each bundle.

app/src/main/java/org/apache/roller/weblogger/pojos/WeblogEntryComment.java
```diff
@@ -130,6 +131,13 @@ public String getUrl() {
     public void setUrl(String url) {
         this.url = url;
     }
+
+    /**
+     * URL of the comment writer when it can be safely rendered as a link.
+     */
+    public String getSafeUrl() {
```
mraible, line 138: Nit: this runs the validator (regex + `DomainValidator` lookup) on every call, and `Comments.jsp` evaluates `#comment.safeUrl` three times per row while `weblog.vm` evaluates `$comment.url` twice per comment. Compute once per render (an `s:set` in the JSP, a local in the macro, or memoize in the wrapper).

app/src/test/java/org/apache/roller/weblogger/ui/rendering/IncomingTrackbackRemovalTest.java
```diff
@@ -0,0 +1,73 @@
...
+                () -> ConfigModel.class.getMethod("getTrackbacksEnabled"));
+        assertThrows(NoSuchMethodException.class,
+                () -> BannedwordslistChecker.class.getMethod(
+                        "checkTrackback",
+                        org.apache.roller.weblogger.pojos.WeblogEntryComment.class));
+    }
+
+    @Test
+    void deploymentAndRuntimeConfigurationDoNotExposeTrackbacks() throws Exception {
+        assertFileDoesNotContain("src/main/webapp/WEB-INF/web.xml", "trackback");
```
mraible, line 65: cwd-relative paths again; surefire sets `project.build.directory` for this module, and no other test in `app/src/test` reads source files relative to the working directory.

app/src/main/java/org/apache/roller/weblogger/pojos/wrapper/WeblogEntryCommentWrapper.java
snoopdave: 🤖Codex: fixed by returning an empty string for null, blank, or invalid wrapped author URLs.

app/src/main/java/org/apache/roller/weblogger/pojos/WeblogEntryComment.java
snoopdave: 🤖Codex: fixed by computing the wrapped URL once and binding the admin-view value once per row.

app/src/main/java/org/apache/roller/weblogger/util/CommentAuthorUrl.java
snoopdave: 🤖Codex: fixed by centralizing posting and rendering on CommentAuthorUrl with URI and IDN support for local, intranet, underscored, and internationalized hosts.

app/src/main/resources/ApplicationResources_de.properties
snoopdave: 🤖Codex: fixed by restoring neutral comment-settings translations in all seven localized bundles.

PR #180, opened 2026-09-06
Update dependencies across project and add JDK 25 to test matrix

mbien:
point release updates within the same major version in most cases


mbien:
ok that worked. checking if we should do webjars and spring point release too.

snoopdave (approved):
LGTM. I can easily roll an RC2 release for this.

PR #182, opened 2026-09-07
Improve installer form initialization and preview navigation

snoopdave:
Addresses the installer and entry-preview findings from Greg’s 6.1.6 RC1 testing.

UI form initialization establishes a session before generating the form salt. Entry previews use a context-relative URL, preserving the editor’s origin. Version-only database upgrades report completion explicitly, and the installer supplies page titles and exception names directly to its templates.

Validation: JDK 11 `mvn -pl app verify` passed (301 tests, zero failures/errors, one existing disabled test). Isolated Tomcat 9.0.64/PostgreSQL checks passed for fresh-session installation at the root context and a version-610 upgrade under `/roller`, including bootstrap and version-616 verification. Preview tests cover HTTP/HTTPS, non-default ports, and both context layouts. Live HTTPS/reverse-proxy preview confirmation on Greg’s deployment remains useful.

### DIFF
--- a/app/src/main/java/org/apache/roller/weblogger/business/startup/DatabaseInstaller.java
+++ b/app/src/main/java/org/apache/roller/weblogger/business/startup/DatabaseInstaller.java
@@ -226,6 +226,8 @@ public class DatabaseInstaller {
 
             log.info("Database is old, beginning upgrade to version "+myVersion);
 
+            boolean schemaChangesRequired = dbversion < 610;
+
             // iterate through each upgrade as needed
             // to add to the upgrade sequence simply add a new "if" statement
             // for whatever version needed and then define a new method upgradeXXX()
@@ -254,6 +256,10 @@ public class DatabaseInstaller {
             // make sure the database version is the exact version
             // we are upgrading too.
             updateDatabaseVersion(con, myVersion);
+            if (!schemaChangesRequired) {
+                successMessage("No table changes were required.");
+            }
+            successMessage("Database version updated to " + myVersion + ".");
 
         } catch (SQLException e) {
             throw new StartupException("ERROR obtaining connection");
--- a/app/src/main/java/org/apache/roller/weblogger/business/startup/DatabaseInstaller.java
+++ b/app/src/main/java/org/apache/roller/weblogger/business/startup/DatabaseInstaller.java
@@ -853,7 +853,8 @@ public class DatabaseInstaller {
     }
 
 
-    private int parseVersionString(String vstring) {
+    // package-private so tests can parse versions exactly as the installer does
+    static int parseVersionString(String vstring) {
         int myversion = 0;
 
         // NOTE: this assumes a maximum of 3 digits for the version number
--- a/app/src/test/java/org/apache/roller/weblogger/business/startup/DatabaseInstallerUpgradeTest.java
+++ b/app/src/test/java/org/apache/roller/weblogger/business/startup/DatabaseInstallerUpgradeTest.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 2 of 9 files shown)


app/src/test/java/org/apache/roller/weblogger/business/startup/DatabaseInstallerUpgradeTest.java
```diff
@@ -0,0 +1,38 @@
...
+        when(rows.next()).thenReturn(true);
+        when(rows.getString(1)).thenReturn("610");
+        when(con.prepareStatement(anyString())).thenReturn(update);
+        DatabaseInstaller installer = new DatabaseInstaller(db, scripts);
+        installer.upgradeDatabase(true);
+        verify(update).setString(1, "616");
+        verify(update).executeUpdate();
+        verifyNoInteractions(scripts);
+        assertTrue(installer.getMessages().stream().anyMatch(m -> m.contains("No table changes were required.")));
+        assertTrue(installer.getMessages().stream().anyMatch(m -> m.contains("Database version updated to 616.")));
```
snoopdave: 🐞GLM Issue: **Important:** The expected version is hardcoded as `"616"` here and in `verify(update).setString(1, "616")` above, which pins this test to the current release number. The next version bump (6.1.6 → 6.1.7) will fail the test even though the behavior it checks is unchanged. Consider deriving the expectation from the same source `DatabaseInstaller` uses (the `ro.version` property in `/roller-version.properties`, parsed the same way `DatabaseInstaller.upgradeDatabase` does), or loosening the assertion to a pattern such as `Database version updated to \d+\.` — that keeps the test asserting the behavior (a version-only upgrade reports completion) rather than a specific release constant.

snoopdave (commented):
PR-Review: 1 inline comment posted.

app/src/test/java/org/apache/roller/weblogger/business/startup/DatabaseInstallerUpgradeTest.java
snoopdave: 🐞GLM Issue: **Resolved** in 4a89f02f45884cfd0e4ed3cdfc4700bbb40a86e0 — the test now derives the expected version from the same source the installer reads (the ro.version property in /roller-version.properties, parsed with DatabaseInstaller.parseVersionString, widened to package-private static so the test parses exactly as production does), instead of pinning the release constant. Full JDK 11 app verify unchanged: 301 tests, 0 failures, 0 errors, 1 skipped.

app/src/main/java/org/apache/roller/weblogger/business/startup/DatabaseInstaller.java
```diff
@@ -254,6 +256,10 @@ public void upgradeDatabase(boolean runScripts) throws StartupException {
             // make sure the database version is the exact version
             // we are upgrading too.
             updateDatabaseVersion(con, myVersion);
+            if (!schemaChangesRequired) {
+                successMessage("No table changes were required.");
+            }
+            successMessage("Database version updated to " + myVersion + ".");
```
mbien, line 262: probably ok for now but this is also a bit brittle.

I think this pattern would be more future proof.

```java
    boolean upgraded = false;

    if(dbversion < xx) {
        upgradeTo...()
        dbversion = yy;
        upgraded = true;
    }
    (...)

   if (!upgraded) { extraMessage(); }
```
But this can be tweaked for the next release.

mbien: gh didn't select the full span, the brittle part is L229 `boolean schemaChangesRequired = dbversion < 610;`.

snoopdave: Thanks — adopted your flag pattern in #184: a `schemaUpgraded` flag set inside each upgrade block, so the "no table changes" message no longer depends on the hardcoded 610. It's behavior-preserving (at that point `dbversion < 610` is true exactly when a block runs), and I added a test for the schema-change path. Folding it into 6.1.6.

PR #184, opened 2026-09-08
Track schema upgrades with a flag instead of a hardcoded version check

snoopdave:
Follow-up to @mbien's review on #182.

`DatabaseInstaller.upgradeDatabase` decided whether to print "No table changes were required." from a hardcoded `boolean schemaChangesRequired = dbversion < 610;`. That constant is the last version with a migration script, so it has to be bumped by hand on every future schema change or the message goes wrong.

This adopts the pattern Michael suggested: a `schemaUpgraded` flag set inside each `if (dbversion < XXX)` upgrade block, with the message printed when `!schemaUpgraded`. The behavior is unchanged — at that point `dbversion` is already `< myVersion` and `>= 310`, so `dbversion < 610` is true exactly when one of the upgrade blocks runs — but the "did a schema step run?" answer is now derived rather than hardcoded, and the guiding comment tells the next maintainer to set the flag.

Also adds a test for the schema-change path (a 520 database upgrading through the 520→610 step, asserting the "no changes" message is suppressed) next to the existing version-only case.

Validation: JDK 11 targeted run green (`DatabaseInstallerUpgradeTest`, 2 tests, 0 failures).

https://claude.ai/code/session_015X69HHQ5XnjRkJP8ymzwFf

### DIFF
--- a/app/src/main/java/org/apache/roller/weblogger/business/startup/DatabaseInstaller.java
+++ b/app/src/main/java/org/apache/roller/weblogger/business/startup/DatabaseInstaller.java
@@ -226,37 +226,46 @@ public class DatabaseInstaller {
 
             log.info("Database is old, beginning upgrade to version "+myVersion);
 
-            boolean schemaChangesRequired = dbversion < 610;
+            // track whether any upgrade step actually ran, so the
+            // "no table changes" message stays correct without a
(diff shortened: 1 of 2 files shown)


mbien (approved):
cool, thanks!

PR #189, opened 2026-09-13
Secure initial Roller setup with one-time operator token

snoopdave:
## Summary

- Require a cryptographically secure, one-time operator token before Roller initial setup, database installation actions, and first-user registration.
- Bind redemption to a browser session, expire it after 60 minutes, and prevent token exposure through URLs or rendered responses.
- Permanently close bootstrap access after the first administrator is committed, including after account deletion or disablement.
- Gate SSO auto-provisioning during bootstrap and document the operator workflow.
- On upgrade, the database upgrade screen is not shown unless there is an actual database upgrade to be done.

## Validation

- `mvn -pl app -DskipTests compile`
- Focused `BootstrapSecurityTest` and `InstallTest`
- `git diff --check`

### New log messages

Here's what the new seucurity code looks like in `roller.log`:

Screenshot 2026-09-13 at 3 24 05 PM

### New security code page 

And here's the new securoty code page:

Screenshot 2026-09-13 at 3 24 31 PM

## Release consideration

If we do an RC4, we should consider including this PR in that release candidate.


app/src/main/java/org/apache/roller/weblogger/ui/core/filters/BootstrapSecurityFilter.java
```diff
@@ -0,0 +1,29 @@
...
+public class BootstrapSecurityFilter implements Filter {
+    public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain) throws IOException, ServletException {
+        HttpServletRequest r = (HttpServletRequest) req;
+        HttpServletResponse p = (HttpServletResponse) res;
+        String uri = r.getRequestURI();
+        boolean tokenPage = uri != null && (uri.endsWith("/bootstrap-token.rol")
+                || uri.endsWith("/bootstrap-token!redeem.rol"));
+        boolean installer = uri != null && (tokenPage || uri.contains("/roller-ui/install/")
+                || uri.endsWith("/roller-ui/register.rol")
+                || uri.endsWith("/roller-ui/register!save.rol")
```
snoopdave, line 19: 🐞Codex Issue: **Blocking:** The gate only matches `/register.rol` and `/register!save.rol`, but Struts dynamic method invocation also exposes URLs such as `/register!execute.rol` and `/register!activate.rol`. Those direct routes can reach the initial registration flow without the token. Gate the resolved register action/method (or reject all register methods while bootstrap is incomplete) and add regression coverage for every allowed method.

snoopdave (commented):
PR-Review: 1 inline blocking comment posted.

snoopdave:
**PR-Review: General Issues**

The following issue needs action but is not tied to one changed line:

- 🐞Codex Issue: **Important:** The new bootstrap security logic has only a completion-state unit test. Add focused tests for successful and replayed redemption, expiry, session binding, invalid methods, concurrent redemption, restart behavior, and direct Struts route variants before merging.

snoopdave:
Need to test this on the 5.2 -> 6.1.6 upgrade, that is the last actual DB schema change

app/src/main/java/org/apache/roller/weblogger/ui/core/filters/BootstrapSecurityFilter.java
```diff
@@ -0,0 +1,29 @@
...
+                || uri.endsWith("/roller-ui/setup.rol"));
+        if (installer && !BootstrapSecurity.isCompleted()) {
```
snoopdave, line 21: Once `isCompleted()` is true this whole block is skipped, so `bootstrap-token.rol` falls through to the action and renders a live-looking token form.

Verified on a fully installed instance (`bootstrap.completed=true` in `roller_properties`): `install.rol` correctly 302s to `/`, but `GET /roller-ui/bootstrap-token.rol` returns **200 with the token `<input>` present**.

It is inert — `digest` is null once setup completes, so any submission gets "The setup token is invalid or has expired" — but it reads against the PR's "permanently close bootstrap access" goal, and an operator who finds that page has no way to tell it is dead. Returning 404, or redirecting to `/` when `isCompleted()`, would match the stated behaviour.

app/src/main/java/org/apache/roller/weblogger/ui/core/RollerContext.java
```diff
@@ -186,6 +191,21 @@ public void contextInitialized(ServletContextEvent sce) {
                 // trigger initialization process
                 weblogger = WebloggerFactory.getWeblogger();
                 weblogger.initialize();
+                try {
+                    org.apache.roller.weblogger.pojos.RuntimeConfigProperty marker =
+                            weblogger.getPropertiesManager().getProperty(BootstrapSecurity.COMPLETION_PROPERTY);
+                    if (marker != null && "true".equalsIgnoreCase(marker.getValue())) {
+                        BootstrapSecurity.complete();
+                    } else if (weblogger.getUserManager().getUserCount() > 0) {
```
snoopdave: This block and the first-admin path in `Register` are the only places that write `bootstrap.completed` and call `complete()` — and this one only runs inside `contextInitialized`.

That leaves a gap on the **upgrade-only** path, which is the common case for an existing site. When the operator redeems the token and completes `install!upgrade.rol` → `install!bootstrap.rol`, the business tier comes up but this block never re-runs, so `completed` stays `false` and no marker is written until the next restart. No first-user registration happens either, since the users already exist.

Observed in that window:
- anonymous `GET /roller-ui/install/install.rol` → 302 to `bootstrap-token.rol`, a page that can no longer be satisfied because `digest` was consumed at redemption — a dead end until Tomcat restarts;
- `bootstrap.completed` absent from `roller_properties`, appearing only after the next restart via this code;
- the redeeming session keeps its 60-minute grant over the installer and `register` actions. That may well be intended, but combined with the above it means "setup finished" and "bootstrap closed" are two different moments.

Calling `complete()` and saving the marker at the end of the successful install/upgrade action would close the gate when setup actually finishes, rather than at the next restart.

app/src/main/java/org/apache/roller/weblogger/ui/struts2/core/BootstrapToken.java
```diff
@@ -0,0 +1,90 @@
...
+        setResponseHeaders();
+        if (!"POST".equalsIgnoreCase(request.getMethod())) {
+            addActionError(getText("installer.bootstrap.postRequired"));
+            return INPUT;
+        }
+        if (BootstrapSecurity.redeem(request, token == null ? null : token.trim())) {
+            LOG.info("One-time setup token accepted; continuing database setup.");
+            return SUCCESS;
+        }
+        addActionError(getText("installer.bootstrap.invalidToken"));
```
snoopdave, line 62: A rejected token produces no distinct log line. The only bootstrap logging is the `LOG.info` in `execute()` saying Roller is waiting for a token — which is emitted on every render of the page, including the re-render after a failure. So repeated failed redemptions are effectively invisible in `roller.log`.

A `LOG.warn` here including `request.getRemoteAddr()` would make probing visible. The token is 256 bits so brute force isn't a realistic threat, but a rejected setup-token submission is exactly the sort of event an operator would want logged.

snoopdave (commented):
🤖Claude: Tested the full upgrade path against a real Apache Roller **5.2.0** binary release (the Nov 2017 `roller-release-5.2.0-standard.zip`) rather than a synthetic schema — installed 5.2.0 on MySQL 8, created a user, weblog and published entry through the UI, then deployed this branch (`20cc99c`) over the same database and upgraded in place.

### What passed

| Check | Result |
|---|---|
| Startup on a 520 database prints the one-time token banner | ✅ |
| `/` redirects to `bootstrap-token.rol` | ✅ |
| Token never appears in the rendered page or in a URL | ✅ |
| Wrong token → "invalid or has expired" | ✅ |
| Correct token → upgrade screen → 520→616 upgrade runs | ✅ |
| **Restart with nothing to upgrade → no upgrade screen, no token issued** | ✅ |
| Second browser session blocked while the first holds the grant | ✅ |
| Token replay in a second session rejected (one-time use) | ✅ |
| Fresh random token per JVM start | ✅ |

The 5.2.0 user, weblog and entry all survived the upgrade, the weblog renders, and `roller.log` is clean.

Three comments inline. One more that can't be line-anchored because it's outside the diff:

### `DatabaseInstaller.tableExists()` (line 820) misdetects tables across schemas on Connector/J 8

```java
ResultSet rs = con.getMetaData().getTables(null, null, "%", null);
```

Connector/J 5.1 read a null catalog as "the current database"; Connector/J 8 reads it as *every* database on the server (`nullCatalogMeansCurrent` now defaults to `false`). On a MySQL server hosting more than one Roller schema, `isCreationRequired()` finds `userrole` / `roller_user` in some *other* database and returns `false`, so a genuinely empty schema is treated as already installed. `WebloggerStartup.prepare()` then falls through to `isUpgradeRequired()`, logs `ERROR setting database version`, and the context dies with `Table '<schema>.roller_properties' doesn't exist` — while Tomcat reports only "One or more listeners failed to start".

This bit while setting up the 5.2.0 instance for this test and took a while to diagnose, since it presents as a datasource or schema problem. Worked around with `nullCatalogMeansCurrent=true` in the JDBC URL; passing `con.getCatalog()` instead of `null` would fix it properly. Not introduced by this PR — raising it because this PR is already reshaping install/upgrade gating in the same class, and the same `tableExists()` is present unchanged back in 5.2.0.

app/src/main/java/org/apache/roller/weblogger/ui/core/filters/BootstrapSecurityFilter.java
mbien: nitpick: i would consider putting 

```
if (BootstrapSecurity.isCompleted()) {
        chain.doFilter(req, res);
        return;
}
```
as fast path to the beginning of the method. (`&& !BootstrapSecurity.isCompleted()` can be removed afterwards). This would also make it clear that the filter becomes a no-op after bootstrap is done. (self documenting code essentially)

PR #191, opened 2026-09-14
Format LDAP comment form values consistently

snoopdave:
Format LDAP comment form field values consistently when redisplaying entered text. Authentication continues to use the original request values.

Adds coverage for first visits, missing fields, ordinary values, punctuation, and literal entity text.

Validation: `mvn -V -ntp install` on JDK 25 — 308 tests, no failures or errors, one skipped.


app/src/main/java/org/apache/roller/weblogger/ui/rendering/plugins/comments/LdapCommentAuthenticator.java
```diff
@@ -84,14 +85,14 @@ public String getHtml(HttpServletRequest request) {
...
-		sb.append(ldapUser + "\">");
+		sb.append(StringEscapeUtils.escapeHtml4(ldapUser)).append("\">");
 		sb.append("</p>");
 		sb.append("<p>");
 		sb.append(messages.getString("comments.ldapAuthenticatorPassword"));
 		sb.append("</p>");
 		sb.append("<p>");
 		sb.append("<input type=\"password\" name=\"ldapPass\" value=\"");
-		sb.append(ldapPass + "\">");
+		sb.append(StringEscapeUtils.escapeHtml4(ldapPass)).append("\">");
```
snoopdave: 🐞Claude Issue: **Blocking:** This repopulates the password field's `value` attribute with the submitted password, so the plaintext credential is written into the rendered page source. From there it is reachable through view-source, the browser's cache and back/forward history, and anything upstream that captures response HTML.

Password inputs should not be repopulated. Drop the `value` attribute for this field entirely:

```java
sb.append("<input type=\"password\" name=\"ldapPass\">");
```

Note that `preservesOrdinaryFormValues` in the new test asserts this round-trip (`sample-pass` appearing in the markup), so that assertion needs updating alongside the fix.

app/src/test/java/org/apache/roller/weblogger/ui/rendering/plugins/comments/LdapCommentAuthenticatorTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private String render(HttpServletRequest request) {
+        try (MockedStatic<CommentAuthenticatorUtils> locales = mockStatic(CommentAuthenticatorUtils.class)) {
+            locales.when(() -> CommentAuthenticatorUtils.getLocale(request)).thenReturn(Locale.ENGLISH);
+            return new LdapCommentAuthenticator().getHtml(request);
+        }
+    }
+
+    private void assertFields(String html, String user, String password) {
+        assertTrue(html.contains("<input name=\"ldapUser\" value=\"" + user + "\">"));
```
snoopdave: 🐞Claude Issue: **Important:** `assertTrue(html.contains(...))` reports only `expected: <true> but was: <false>` when it fails — no expected value, no actual HTML, and no indication which of the two fields broke. I confirmed this by running the suite against the pre-change behaviour; the failure output gives a maintainer nothing to work with.

Use the message overload, or compare extracted values directly:

```java
assertTrue(html.contains(expected), () -> "expected " + expected + " in " + html);
```

snoopdave (commented):
PR-Review: 2 inline comments posted.

app/src/main/java/org/apache/roller/weblogger/ui/rendering/plugins/comments/LdapCommentAuthenticator.java
snoopdave: 🤖Claude: fixed by removing the `value` attribute from the password input entirely, so the submitted password is no longer written into the rendered markup. The now-unused `ldapPass`/`ldapPassTemp` locals in `getHtml` were dropped with it; `authenticate()` still reads the parameter straight off the request, so sign-in is unaffected. Added `omitsSubmittedPasswordFromMarkup` as a regression test, and updated `preservesOrdinaryFormValues` (and the shared `assertFields` helper) which previously asserted the round-trip. Verified red/green on JDK 11: restoring the `value` attribute fails the new test.

app/src/test/java/org/apache/roller/weblogger/ui/rendering/plugins/comments/LdapCommentAuthenticatorTest.java
snoopdave: 🤖Claude: fixed by giving both assertions the supplier-message overload, so a failure now names the expected markup and prints the actual HTML instead of just `expected: <true> but was: <false>`. Confirmed against a deliberately broken build — the failure now reads `password written into markup: <p>...<input type="password" name="ldapPass" value="sample-pass"></p>`, which identifies the offending field directly.

snoopdave:
🤖Claude: Pushed `18a840ae5`, which widens this PR slightly beyond its title — flagging it so reviewers are not surprised.

**What changed**

The password input is now rendered with no `value` attribute at all:

```java
sb.append("<input type=\"password\" name=\"ldapPass\">");
```

Previously it was rendered with the submitted password as its value. Escaping that value made the markup well-formed, but the credential was still being written into the response body, where it is reachable through view-source, browser cache and back/forward history, and anything upstream that records response HTML. Not returning it at all is the stronger position, and it is what browsers and password managers expect from a password field.

The locals that carried the password through `getHtml` were removed with it. **`authenticate()` is unaffected** — it reads `request.getParameter("ldapPass")` directly, so signing in behaves exactly as before. The only user-visible change is that a commenter who mistypes their password retypes it rather than finding the field pre-filled, which is normal for a password input.

**Tests**

- `omitsSubmittedPasswordFromMarkup` is new and covers the rendering.
- `preservesOrdinaryFormValues` and the shared `assertFields` helper previously asserted the password round-trip, so both were updated.
- Assertions now use the supplier-message overload: a failure names the expected markup and prints the actual HTML, instead of reporting only `expected: <true> but was: <false>`.

Verified red/green on JDK 11 — restoring the `value` attribute fails the new test. All 15 comment-related tests pass.

PR #193, opened 2026-09-19
No single quotes in media file insert url on mediaFileAdd!save.rol

gregh3269:
The uploaded confirmation page has a link "Create new weblog post" where it adds the image to a new weblog post, but uses single quotes and the image does not show correctly after saving/publishing (but should).  Add media file from the New Entry page works, and uses double quotes.

Sorry, thought this was fixed by Matt's pr.


mbien:
this looks ok but we should probably filter through the code base and check all single quote usages to avoid having to update those one by one.

gregh3269:
I looked, could not find any more struts actions that adds html code.

Thanks.

